PR #22, opened 2013-12-31
Added Wildcard Subdomain Clarification

cillosis:
As per our conversation in the #vagrant IRC, I've added a note to the README to provide a small amount of clarification for others attempting to use wildcard subdomains with hostnames other than vagrant.dev. I would have updated more, but am still a little unclear as to how the code functions as I am not intimately familiar with Ruby.


itsazzad:
will for example  `www.example.dev` not work?

PR #62, opened 2014-04-19
Add support for arbitrary record types

benlangfeld:


bexelbie:
@benlangfeld Can you rebase this and let's look at trying to get this committed.

benlangfeld:
@bexelbie Is this project alive again? If I rebase this is it likely to get merged?

bexelbie:
@benlangfeld alive and kicking :) We had our first maintainers meeting and have a basic Roadmap to get us rolling again

benlangfeld:
@bexelbie Rebased, but this modifies CNAME support and removes reverse DNS. I'm not convinced either were implemented adequately and were done so in a way that makes generic record type support hard. This needs a ground-up rethink of what landrush is supposed to be, along the lines of the notes above.

I've made my preference for the approach clear. If you're in agreement, then all that remains is to decide the API Landrush is to provide for manual record insertion and which records should be inserted automatically, both to generate a BIND zone file. That this functionality and the DNS server are coupled has over time encouraged strange behaviours and a very loose specification of intent.

bexelbie:
@benlangfeld I will take some time to re-read the history of this PR before I comment.  I am updating to let you know that I saw your update.

hferentschik:
@benlangfeld, even worse. I merged now also my changes to make Landrush work on Windows. They don't directly affect your code, but since I touched a lot of files, there are plenty of merge issues. Not sure if you to try to sort them out. If so, it would be nice if you also create a issue for this and add the issue key to the commit messages. 

The new issue could also serve as a discussion ground. AFAIU you also think that we first need to define the overall goal. For me it is not quite clear yet which benefit supporting BIND zone files has? What do we really gain here, compared to the simple config we have now? Could additional DNS entry types also supported by sticking to a simple JSON notation?

README.md
```diff
@@ -47,6 +47,7 @@ If you are using a multi-machine `Vagrantfile`, configure this inside each of yo
 You can add static host entries to the DNS server in your `Vagrantfile` like so:
 
     config.landrush.host 'myhost.example.com', '1.2.3.4'
+    config.landrush.host '_sip._udp.example.com', [1, 0, 5060, 'myhost.example.com'], 'srv'
```
hferentschik: I  guess some explanation would be needed here what this actually means. They make no sense to me atm. I guess they are some sort of BIND syntax.

lib/landrush.rb
```diff
@@ -12,6 +14,10 @@ def self.working_dir
   def self.working_dir=(working_dir)
     @working_dir = Pathname(working_dir).tap(&:mkpath)
   end
+
+  def self.config
+    @config ||= PStore.new(working_dir.join('config.pstore'))
```
hferentschik: That means the config is not a pstore which is not human readable anymore, right? I think this is a bit unfortunate. Having the ability to easily view/edit the configuration is imo a nice thing.

benlangfeld:
@hferentschik I don't use Landrush currently; in fact, right now I have no need for anything landrush-like. Given my other project load, I probably won't get back to this issue. If I ever do have a need in the future and the point of this PR becomes important, I will probably build it out separately from Landrush to avoid integration effort.

hferentschik:
> I don't use Landrush currently; in fact, right now I have no need for anything landrush-like. 

Fair enough.

PR #94, opened 2015-01-06
update vagrant version to 1.7.1

d-snp:
Hi, I couldn't get it to bundle because the vagrant you locked on had some kind of incompatibility with my bundler version, so I updated the vagrant version, was a small fix getting the specs to pass.


phinze:
Cool - this makes sense. Thanks!

PR #95, opened 2015-01-06
Rubydns 0 9 4

d-snp:
This pull request upgrades to the latest version of rubydns, no idea if it's necessary, but it seems 0.7.3 is a bit buggy. All specs pass, I'm not sure if they cover enough for you to be sure this upgrade is good. I haven't run it in production yet.

The specs did become very chatty, RubyDNS is logging all kinds of information.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,12 +3,16 @@ source 'https://rubygems.org'
 # Can't use `gemspec` to pull in dependencies, because the landrush gem needs
 # to be in the :plugins group for Vagrant to detect and load it in development
 
-gem 'rubydns', '0.7.3'
+gem 'rubydns', '0.9.4'
 gem 'rake'
 
 # Vagrant's special group
 group :plugins do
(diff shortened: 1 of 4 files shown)


d-snp:
Also, it introduces an explicit dependency on rexec, which has been made obsolete by process-daemon. I didn't have much time to figure out how to port your code to that gem so I just let it remain dependent on RExec.

phinze:
This looks good. I'll see about porting over to process-daemon. Can you rebase this since I just merged #94?

d-snp:
It's already based on #94, I think it should merge cleanly, or does github warn you?

phinze:
Oh cool. I'm probably just a dummy. :grinning:

phinze:
Yup, dummy status confirmed!

d-snp:
no, I was just totally cheating by assuming you were going to merge #94 without modification ;)

PR #96, opened 2015-01-08
Command-line improvements for `vagrant landrush`

ahpook:
There are two commits here, which you should be able to take independently if you want:

1. Previously, running 'vagrant landrush' without any actions would produce an ugly error. Now it just defaults to 'help'.
2. Previously, there was no way to manipulate the DNS database without mucking with it directly or bringing new hosts up/down. I added `set` and `rm` commands to manipulate it through the Store API.

This was based on an afternoon of poking through it so might be completely wrong. I noticed there aren't tests to cover the command line actions and did not dive into adding any here, but at least no existing tests broke!

### DIFF
--- a/lib/landrush/command.rb
+++ b/lib/landrush/command.rb
@@ -9,7 +9,7 @@ module Landrush
     def execute
       ARGV.shift # flush landrush from ARGV, RExec wants to use it for daemon commands
 
-      command = ARGV.first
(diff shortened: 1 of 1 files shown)


phinze:
This looks great - thanks!

ahpook:
:+1:

PR #97, opened 2015-01-08
include rexec in gem dependencies

d-snp:
I didn't know that you use rexec in the gem itself as well, not just the tests. This adds the dependency. I've tested it and with this fix the gem seems to work when installed as a vagrant plugin.


phinze:
:cake:

PR #21, opened 2013-12-18
adds reverse dns support

chadh:
This patch adds support for reverse dns queries (e.g. `host 10.250.0.11`)

This works for me, but I'm a novice ruby coder and may be overlooking something.  In particular, I had to remove the "redirect" in `Store.hosts.get()` for my implementation to work.

PR #98, opened 2015-01-18
Add support for IN::PTR records.

metcalfc:
Building on the work done by @chadh in #21 I believe this should cover this feature. I also tried to clean it up a bit. So as to not add another gem dependency I used `ipaddr`. Its not as pretty but it gets the job done. I added two conditions to the redirect logic which seems to work. I added a test for the ptr record. Test pass.


phinze:
Thanks @metcalfc - this looks good!

PR #99, opened 2015-01-20
Add support for CNAME records.

metcalfc:
#96 added set and got me thinking about what I'd use that for. It quickly lead to CNAME and here's the patch that implements a simple CNAME lookup. Basically if a hostname exists as a value and a key this should work. You'll get a CNAME and the resulting A record for the corresponding entry.

Tests pass. No additional tests added per #96 it wasn't clear how to best add them. Happy to if there are suggestions.

-----

With support for set on the command line its trival for a user to:

  vagrant landrush set some.host.com some.other.com

Seems like we should check for possible CNAMEs in the current config.

### DIFF
--- a/lib/landrush/server.rb
+++ b/lib/landrush/server.rb
@@ -47,6 +47,17 @@ module Landrush
       super
     end
 
+    def self.check_a_record (host, transaction)
+      value = Store.hosts.get(host)
+      if (IPAddr.new(value) rescue nil)
+        name = transaction.name =~ /#{host}/ ? transaction.name : host
+        transaction.respond!(value, {:ttl => 0, :name => name})
+      else
+        transaction.respond!(Name.create(value), resource_class: IN::CNAME, ttl: 0)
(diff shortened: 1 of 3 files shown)


phinze:
Hey this looks great! Coupla quick notes:

 * This could probably use some test coverage
 * We have some [janky-ass fake CNAME support](https://github.com/phinze/landrush/commit/a6a01dc83aafcd5506abb44d7fc98428dfc73bb9) jacked into the `Store` - we should probably pull that out at the same time this lands.

metcalfc:
Sounds good. I'll try and get this done this week. Pulling most of the redirect will really simplify the logic. :boom:

metcalfc:
Killed the redirect logic and added a test. All tests pass. Let me know if there is anything else.

phinze:
This is awesome. Thank you!

metcalfc:
Woot! When you have a moment can you cut a release? Thanks!

Issue #9, opened 2013-07-03
iptables rules lost on machine reboot


phinze:
so if you reboot your machine without doing `halt` / `up`, your DNS will revert to non-landrush mode.

obviously the workaround is to do the `halt` / `up` cycle, but that's not very swag. :bowtie:

phinze:
This is fixed for reload in #101 - still not handled for guest-initiated reboot, so leaving open for that.

PR #101, opened 2015-01-24
finally support reload!

phinze:
just needed to figure out how action_hook worked in a little more detail

now the landrush actions should get injected into the builder for all
the relevant commands


njam:
Thank you!

Also closes #9?

PR #106, opened 2015-04-22
Update apt indices before installing

njam:


njam:
@phinze fyi, I've added updating of apt indices before running `install`, as recommended [by travis](http://docs.travis-ci.com/user/build-configuration/#Installing-Packages-Using-apt).

PR #125, opened 2015-09-03
Issue #177 Allow multiple TLD and resolver configs

mheiges:
This PR allows `landrush.tld` to take an array of TLD values. Landrush will create OS X resolver configurations for each TLD. There is no change to the existing syntax for assigning a single value.

An example Vagrant configuration would look like

```
config.landrush.tld = ['vm.example.com', 'vm.test.org']
config.landrush.host 'www.vm.example.com', HOST_IP
config.landrush.host 'www.vm.test.org', HOST_IP
```

An example benefit for multiple TLD comes when running multiple websites on a single virtual machine using Apache/Nginx name-based vhosting. That is, a VM can host www.vm.example.com and www.vm.test.org and clients can resolve both domains.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,10 @@ If you would like to configure your guests to be accessible from the host as sub
 
     config.landrush.tld = 'vm'
 
+Multiple TLD can also be provided as an array. On OS X, a resolver configuration will be created for each TLD.
+
+    config.landrush.tld = ['vm', 'dev']
+
 Note that from the __host__, you will only be able to access subdomains of your configured TLD by default- so wildcard subdomains only apply to that space. For the __guest__, wildcard subdomains work for anything.
 
 ### Unmatched Queries
--- a/lib/landrush/action/common.rb
+++ b/lib/landrush/action/common.rb
@@ -71,7 +71,7 @@ module Landrush
           return machine.config.vm.hostname
         end
 
-        "#{Pathname.pwd.basename}.#{config.tld}"
+        "#{Pathname.pwd.basename}.#{config.tld[0]}"
       end
 
       def enabled?
--- a/lib/landrush/action/setup.rb
+++ b/lib/landrush/action/setup.rb
@@ -28,7 +28,9 @@ module Landrush
       end
 
       def setup_host_resolver
-        ResolverConfig.new(env).ensure_config_exists!
+        config.tld.each { |tld|
+          ResolverConfig.new(env, tld).ensure_config_exists!
+        }
       end
 
       def add_prerequisite_network_interface
@@ -64,8 +66,15 @@ module Landrush
         ip_address = machine.config.landrush.host_ip_address ||
                      machine.guest.capability(:read_host_visible_ip_address)
 
-        if not machine_hostname.match(config.tld)
-          log :error, "hostname #{machine_hostname} does not match the configured TLD: #{config.tld}"
+        tld_match = false
+        config.tld.each { |tld|
+          if machine_hostname.match(tld)
+            tld_match = true
+            break
+          end
+        }
+        if !tld_match
+          log :error, "hostname #{machine_hostname} does not match any of the configured TLD: #{config.tld}"
           log :error, "You will not be able to access #{machine_hostname} from the host"
         end
 
--- a/lib/landrush/config.rb
+++ b/lib/landrush/config.rb
@@ -36,6 +36,10 @@ module Landrush
       !!@enabled
     end
 
+    def tld
+      Array(@tld)
+    end
+
     def guest_redirect_dns?
       @guest_redirect_dns
     end
--- a/lib/landrush/resolver_config.rb
+++ b/lib/landrush/resolver_config.rb
@@ -18,8 +18,9 @@ module Landrush
       self.class.config_dir
     end
 
-    def initialize(env={})
+    def initialize(env={}, tld)
       @env = env
+      @tld = tld
     end
 
     def info(msg)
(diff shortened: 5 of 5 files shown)


hferentschik:
I pushed a rebased version of this pull request to here - https://github.com/hferentschik/landrush/tree/issue-177

README.md
```diff
@@ -58,6 +58,10 @@ If you would like to configure your guests to be accessible from the host as sub
 
     config.landrush.tld = 'vm'
 
+Multiple TLD can also be provided as an array. On OS X, a resolver configuration will be created for each TLD.
```
hferentschik: So what's the purpose here really? You need visibility of these TLD on the hosts, right? And as it stands now, this will only work for OS X out of the box. In fact even making the single TLD work out of the box works only for OS X atm.

lib/landrush/action/common.rb
```diff
@@ -71,7 +71,7 @@ def read_machine_hostname
           return machine.config.vm.hostname
         end
 
-        "#{Pathname.pwd.basename}.#{config.tld}"
+        "#{Pathname.pwd.basename}.#{config.tld[0]}"
```
hferentschik: So why the first element of the array? Don't you have to take any action for potential others? This looks a bit fishy to me and makes me wonder whether what one really wants is a config option which lets you define host names to be made visible on the host. There is _config.landrush.host_, could that not be used? One would just make sure that one calls the _ResolverConfig_ for each configured host. WDYT?

hferentschik:
On a tangent, any idea why the automatic host configuration is only applied for OS X? Obviously due to _/etc/resolver_ its a bit easier, but one could also run the required _dnsmasq_ commands for Linux, right? Same for Windows really. There I am working on a prototype, but it is work in progress.

lib/landrush/action/common.rb
mheiges: My understanding is that method sets a default DNS hostname entry. If `config.vm.hostname` is not defined in the Vagrantfile it generates one from the `landrush.tld`. A single default hostname entry is suitable for my use cases so, to keep code changes to a minimum, I just use the first tld.

In my Vagrantfile I use

```
config.landrush.tld = [ 'one.org', 'two.org' ]
config.landrush.host 'one.org'
config.landrush.host 'two.org'
```

So, yes, I use `config.landrush.host` but I only define hosts at the tld level and leverage landrush's wildcard subdomains. This allows me to configure apache/nginx namebased virtual hosts: a.one.org, b.one.org, a.two.org without having to manually manage and reload individual hostnames in the Vagrantfile.

README.md
mheiges: right

mheiges:
auto config for Windows and Linux is outside my expertise but certainly would be a welcomed feature.

lib/landrush/action/common.rb
hferentschik: > So, yes, I use config.landrush.host but I only define hosts at the tld level and leverage landrush's wildcard subdomains. This allows me to configure apache/nginx namebased virtual hosts: a.one.org, b.one.org, a.two.org without having to manually manage and reload individual hostnames in the Vagrantfile.

thanks for the clarification

hferentschik: >    config.landrush.tld = [ 'one.org', 'two.org' ]
>    config.landrush.host 'one.org'
>    config.landrush.host 'two.org'

That's what I mean, you still need 'config.landrush.host'. Would you not prefer something like:

    config.landrush.host 'two.org' true

Where the boolean specifies whether the host entry is visible on the host. Really all this means is in the case of OS X that there will be an entry 'two.org' in _/etc/resolver_.

You only use _config.landrush.tld_, since atm that is the config option which will generate the required config, right? 

Wildcard subdomains come for "free". What I mean is that:

    config.landrush.tld = 'one.org'
    config.landrush.host 'one.org'
    config.landrush.host 'two.org'

works just fine with subdomains, if you just do the following:

    sudo cp -r /etc/resolver/one.org /etc/resolver/two.org

mheiges: >     config.landrush.host 'two.org' true

So I would _still_ need 'config.landrush.host'? But now with a boolean that is less self-documenting.

This is more clear to me (and perhaps only me :-) ) because simply builds on existing syntax:

```
config.landrush.tld = [ 'one.org', 'two.org', ... ]
config.landrush.tld.each do |tld|
  config.landrush.host tld
end
```

> sudo cp -r /etc/resolver/one.org /etc/resolver/two.org

Sure, there are many ways to achieve the end goal and going outside landrush to manage extra TLDs is one of them, but that seems less clean to me.

th3fallen:
any ideas when this may be merged?

hferentschik:
> any ideas when this may be merged?

I was a bit on a back-burner recently. I am also still wondering whether the suggested approach is the way to go. Finally got around finishing my pull request for Windows network configuration. Looking at this shortly again...

th3fallen:
> I was a bit on a back-burner recently. I am also still wondering whether the suggested approach is the way to go. Finally got around finishing my pull request for Windows network configuration. Looking at this shortly again...

Awesome super excited about it

KuiKui:
Hi ! Any ideas when this may be merged ?

goodwillcoding:
@hferentschik ... any ideas on when this is will make it in :-D ? Still a very useful feature? Do you need any help merging and publishing the GEM? Or is lack of linux support an issue here?

/cc: @mheiges

denizdogan:
It's been more than two years, so I just want to ask if this project is still alive and if this PR is likely to ever be merged?

hferentschik:
Superseded by pull request #332 which merges this approach onto master.

PR #140, opened 2015-12-05
(work in progress) Introducing rubocop

fh:


njam:
:+1: 
Maybe run `rubocop` first in `Rakefile`, because it fails faster?

fh:
@njam Good idea! However, looking at the PR, I realized that there are two more commits in here that should be a separate PR. 

Moved the rubocop intro to https://github.com/phinze/landrush/pull/142, closing this PR.

PR #142, opened 2015-12-06
Introducing rubocop

fh:


njam:
Travis:
> NoMethodError: undefined method `spec' for nil:NilClass

I had this on some other travis ruby builds - seems in *some* cases Travis currently uses a different Ubuntu 12.04 environment, which leads to this error.
Workraround could be to add `dist: trusty` to the travis config, which will make it use Ubuntu 14.04. Might be more future-compatible anyway…

fh:
@njam Yes, seems something along those lines. I created a second PR for random travis improvements, and opted to first wait for that one to be merged before diving in to travis debugging foobar :) 
An update to trusty sure is a good idea, just was not 100% sure if thats the whole underlying problem to my PR.

references https://github.com/phinze/landrush/pull/141

fh:
@njam Seems like rebasing to the new docker-based travis setup fixed it. On that note, since you did the initial travis setup: Is there a dedicated reason to use vagrant 1.7.1 in the setup? The newest tagged on github (1.7.4) would enable us to use the newest bundler version.

njam:
1.7.1 was probably just the newest stable version back then. Agree we can go with 1.7.4!

Changes in this PR look good.

PR #143, opened 2015-12-06
Add development notes

njam:


njam:
@fh wdyt?

fh:
:+1: Maybe adjust https://github.com/phinze/landrush/blob/master/.travis.yml#L20, too, while you are at it - the new version requirement seems to be [`<= 1.10.5`](https://github.com/mitchellh/vagrant/blob/v1.7.4/vagrant.gemspec#L18).

lgtm otherwise.

njam:
Oh missed that!

Like this?

fh:
good to merge. :)

Issue #136, opened 2015-12-01
Acceptance CI tests


njam:
Would be great is to have some acceptance tests, because a lot of the problems are dependent on the host or guest OS' particularities. Of course those can't run on travis.

Is there another free CI service that allows to run Virtualbox VMs?
Otherwise we could look into running a jenkins on some dedicated hosting.

fh:
@njam This is rather complicated - the basic reason here is that most (at least all I am aware of) run their tests in a container of some sort, and you cant start a vagrant virtualization within an already virtualized env). At work, we currently develop some sort of "CI" for vagrant running on a normal root server - while this could technically work, it obviously wont be free.

As a first step, we maybe could still shell out those tests to be able to work in the respective dev env, and then later test ideas how to run them in a CI env?

fh:
Just ran into https://github.com/mitchellh/vagrant-spec - maybe thats worth looking into?

njam:
Thanks for the link! I started to draft two different options:

**cucumber: https://github.com/phinze/landrush/pull/144**
- + Good docu, community, support
- + Can run scenarios with different environments (providers, boxes) - "Scenario Outline"
- - Test isolation needs to be done manually after each test (`vagrant destroy`) 

**vagrant-spec: https://github.com/phinze/landrush/pull/145**
- + Explicitly for vagrant, robust test isolation
- - Outdated and only little docu or examples
- - Can't find good way to run test in different environments

I would rather go with *cucumber*. What do you think? Or we should go a completely different way?

fh:
Wow, using cucumber here is a cool idea. Totally up for giving it a shot.

fh:
@njam So, finally was able to resolve my vagrant issues, and gave both approaches a shot. However, the cucumber version still seems not to be fully working, some test cases fail (details in the pr).

I agree with your assessment on vagrant-spec, though, it seems to be not that well suited for our cases here. I would suggest to close the vagrant-spec pr, and focus on cucumber, to get it in a better/fully running state. Not sure how to get this running within travis since a vm cant run in a vm, but maybe we can come up with another solution for a CI env. 

What do you think, how should we proceed from here?

njam:
Closing in favor of https://github.com/vagrant-landrush/landrush/pull/144

hferentschik:
Merged pull request #182.

PR #144, opened 2015-12-06
Acceptance tests: cucumber

njam:
Run with:
```
bundle exec rake features
```

Todo:
Upgrade to Aruba 0.13, and add:
```rb
    Aruba.configure do |config|
      config.activate_announcer_on_command_failure = [:stdout, :stderr]
    end
```

### DIFF
--- a/features/step_definitions/dns.rb
+++ b/features/step_definitions/dns.rb
@@ -1,0 +1,20 @@
+Then(/^the hostname "([^"]+)" should resolve to "([^"]+)" on the internal DNS server$/) do |host, ip|
+  run("dig +short @localhost -p 10053 '#{host}' A")
+  expect(last_command_started).to have_output(/^#{ip}$/)
+end
+
+Then(/^the hostname "([^"]+)" should resolve to "([^"]+)" on the host$/) do |host, ip|
+  if RbConfig::CONFIG['host_os'] =~ /darwin|mac os/
+    run('sudo killall -HUP mDNSResponder')
+    run("ping -c 1 -t 1 '#{host}'")
+    expect(last_command_started).to have_output(/\(#{ip}\)/)
+  else
+    run("dig +short '#{host}' A")
+    expect(last_command_started).to have_output(/^#{ip}$/)
+  end
+end
+
+Then(/^the hostname "([^"]+)" should resolve to "([^"]+)" on the guest/) do |host, ip|
+  run("vagrant ssh -c \"dig +short '#{host}' A\"")
+  expect(last_command_started).to have_output(/^#{ip}$/)
+end
--- a/Gemfile
+++ b/Gemfile
@@ -24,4 +24,7 @@ group :development do
   gem 'byebug'
   gem 'mocha'
   gem 'minitest'
+  gem 'cucumber', '~> 2.1'
+  gem 'aruba', '~> 0.13'
+  gem 'komenda', '~> 0.0.4'
 end
--- a/README.md
+++ b/README.md
@@ -118,6 +118,10 @@ Run the test suite:
 
     bundle exec rake
 
+Run cucumber/aruba acceptance tests:
+
+    bundle exec cucumber
+
 Run the vagrant binary with the landrush plugin loaded from your local source code:
 
     bundle exec vagrant landrush <command>
--- a/Rakefile
+++ b/Rakefile
@@ -1,6 +1,7 @@
 require 'bundler/gem_tasks'
 require 'rake/testtask'
 require 'rubocop/rake_task'
+require 'cucumber/rake/task'
 
 Rake::TestTask.new do |t|
   t.pattern = 'test/**/*_test.rb'
@@ -17,3 +18,5 @@ task :generate_diagrams do
 end
 
 RuboCop::RakeTask.new
+
+Cucumber::Rake::Task.new(:features)
--- a/features/dns_resolution.feature
+++ b/features/dns_resolution.feature
@@ -1,0 +1,24 @@
+Feature: dns_resolution
+  Landrush should make a virtual machine's IP address DNS-resolvable.
+
+  Scenario Outline: booting a box
+    Given a file named "Vagrantfile" with:
+    """
+    Vagrant.configure('2') do |config|
+      config.vm.box = '<box>'
+      config.vm.hostname = 'my-host.my-tld'
+      config.vm.network :private_network, ip: '10.10.10.123'
+
+      config.landrush.enabled = true
+      config.landrush.tld = 'my-tld'
+    end
+    """
+    When I successfully run `bundle exec vagrant up --provider <provider>`
+    Then the hostname "my-host.my-tld" should resolve to "10.10.10.123" on the internal DNS server
+    And the hostname "my-host.my-tld" should resolve to "10.10.10.123" on the host
+    And the hostname "my-host.my-tld" should resolve to "10.10.10.123" on the guest
+
+    Examples:
+      | box             | provider   |
+      | debian/jessie64 | virtualbox |
+      | ubuntu/wily64   | virtualbox |
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -1,0 +1,14 @@
+require 'aruba/cucumber'
+require 'komenda'
+
+Aruba.configure do |config|
+  config.exit_timeout = 300
+  config.activate_announcer_on_command_failure = [:stdout, :stderr]
+end
+
+After do |_scenario|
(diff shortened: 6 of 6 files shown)


fh:
@njam Gave this a try - having some issues here. First run seemed to stall at some point with an error message:

```
    Examples: 
      | debian/wheezy64 | virtualbox |
      expected "bundle exec vagrant up --provider virtualbox" to be successfully executed (RSpec::Expectations::ExpectationNotMetError)
      features/dns_resolution.feature:21:in `When I successfully run `bundle exec vagrant up --provider virtualbox`'
      features/dns_resolution.feature:16:in `When I successfully run `bundle exec vagrant up --provider <provider>`'
      |


^C
Exiting... Interrupt again to exit immediately.
```

Cancelling this was apparently a mistake, there seems to be "somewhere" a vagrant lockfile or something similar around (cant find it, though):

```
    Examples: 
      | debian/wheezy64 | virtualbox |
      Unable to destroy vagrant environment:
      You appear to be running Vagrant outside of the official installers.
      Note that the installers are what ensure that Vagrant has all required
      dependencies, and Vagrant assumes that these dependencies exist. By
      running outside of the installer environment, Vagrant may not function
      properly. To remove this warning, install Vagrant using one of the
      official packages from vagrantup.com.
      An action 'destroy' was attempted on the machine 'default',
      but another process is already executing an action on the machine.
      Vagrant locks each machine for access by only one process at a time.
      Please wait until the other Vagrant process finishes modifying this
      machine, then try again.
      
      If you believe this message is in error, please check the process
      listing for any "ruby" or "vagrant" processes and kill them. Then
      try again.
       (RuntimeError)
      ./features/support/env.rb:10:in `block (2 levels) in <top (required)>'
      features/dns_resolution.feature:21:in `When I successfully run `bundle exec vagrant up --provider virtualbox`'
      features/dns_resolution.feature:16:in `When I successfully run `bundle exec vagrant up --provider <provider>`'
```

njam:
Running the test can obviously take quite some time - because it's starting and destroying boxes.

But you mentioned that in your case it really stalled. Unfortunately I noticed that cucumber/aruba's debugging output is not too useful. For example if a command fails they don't show the command's output… I will look into trying to change that.
Are you sure the test was stalled? I.e. no CPU was used anymore?

When you cancelled the test the cleanup behaviour was triggered (`vagrant destroy`). But vagrant failed because it detected another vagrant command still running (probably `vagrant  up`). You need to clean this up manually by running for example `pkill -f vagrant`.

fh:
@njam Yes, tried those, to no avail, also `ps aux|grep vagrant` or `|grep ruby` didnt show anything - thats why I suspected a lockfile. Will debug this further. :)

njam:
Hmhmhmhm… I was experiencing troubles with this "locking" mechanism in previous Vagrant versions, but not anymore with the current 1.7.4. Just in case you're on an older version.

fh:
Fixed this with rebooting (wtf?), but still having some issues:

```
    Examples: 
      | debian/wheezy64 | virtualbox |
      expected "bundle exec vagrant up --provider virtualbox" to be successfully executed (RSpec::Expectations::ExpectationNotMetError)
      features/dns_resolution.feature:21:in `When I successfully run `bundle exec vagrant up --provider virtualbox`'
      features/dns_resolution.feature:16:in `When I successfully run `bundle exec vagrant up --provider <provider>`'
      | debian/jessie64 | virtualbox |
      expected "bundle exec vagrant up --provider virtualbox" to be successfully executed (RSpec::Expectations::ExpectationNotMetError)
      features/dns_resolution.feature:22:in `When I successfully run `bundle exec vagrant up --provider virtualbox`'
      features/dns_resolution.feature:16:in `When I successfully run `bundle exec vagrant up --provider <provider>`'

Failing Scenarios:
cucumber features/dns_resolution.feature:21 # Scenario Outline: booting a box, Examples (#1)
cucumber features/dns_resolution.feature:22 # Scenario Outline: booting a box, Examples (#2)

2 scenarios (2 failed)
6 steps (2 failed, 2 skipped, 2 passed)
```

njam:
Hmm, yeah it's a pity cucumber/aruba doesn't show the output of a failed command…
It looks like the `vagrant up` already fails.
Could you try manually on a console?

Create a `Vagrantfile`:
```
    Vagrant.configure('2') do |config|
      config.vm.box = 'debian/wheezy64'
      config.vm.hostname = 'my-host.my-tld'
      config.vm.network :private_network, ip: '10.10.10.123'

      config.landrush.tld = 'my-tld'
    end
```

Then start it up:
```
bundle exec vagrant up --provider virtualbox
```

fh:
@njam Lol - when I tried this just now, everything suddenly works, including the originally failing tests. The only change I did yesterday was to codesign the ruby binary here in osx in between to finally get rid of the osx security alert during every vagrant start. I guess this was for some reason affecting the tests, too - cant reproduce it for now.

In any case: lgtm. As soon as I am able to reproduce the issue, I'll debug further, but as of now, the tests work completely. :dancers:

njam:
I opened a ticket on aruba to display the command's output on failure: https://github.com/cucumber/aruba/pull/335

I've adjusted the tests a bit, adding checks for DNS resolution with the custom resolver and also the host's and the guest's resolver.

I'm not sure why, but I have to flush OSX' DNS resolver for the tests to succeed. I hope this is not a general problem, but something with my setup…

@fh wdyt?

njam:
I've added some basic acceptance tests that cover the DNS resolution functionality of landrush with a virtualbox VM. You can run them with:
```
bundle exec cucumber
```

They succeed for "debian/jessie64", but fail for "ubuntu/wily64". This is actually a problem of our IP detection (on wily we pick up the address of an additional LXC interface that is present). This could be fixed by https://github.com/phinze/landrush/pull/122.

I would suggest to merge something like what is on this branch, so we can test other branches with it.

@fh @ahpook @strzibny @bexelbie wdyt?

features/step_definitions/dns.rb
```diff
@@ -0,0 +1,20 @@
+Then(/^the hostname "([^"]+)" should resolve to "([^"]+)" on the internal DNS server$/) do |host, ip|
```
hferentschik: I like the overall idea of cucumber, but I am not so sure about these steps. They rely on Unix tools like  _dig_ and _killall_. This won't work on Windows. I spend quite a some effort to a) get Landrush to work on Windows and b) make it possible to develop on Windows including running the test suite. This would break this again.

If we can find a way to make this cucumber tests work w/o relying on Unix tools I am all up for it. For example instead of using _dig_ one could just use Ruby's Resolv:DNS class.

hferentschik:
I am tempted to close this pull request in this form. @njam WDYT?

njam:
I think it would be great to have acceptance tests for landrush.
Agree with your concerns.
I will adjust this branch to be cross-platform compatible very soon, wdyt?

hferentschik:
> I think it would be great to have acceptance tests for landrush.

+1, btw have you looked at [vagrant-spec](https://github.com/mitchellh/vagrant-spec). I used it for other vagrant plugins as well. It is not perfect, in particular I dislike that the "core" tests are always visible. On the other hand it is quite intuitive and offers black box testing of your plugin. IMO with less effort than cucumber.

> Agree with your concerns.

Cool

> I will adjust this branch to be cross-platform compatible very soon, wdyt?

Sure. It would be nice to see a couple of more tests then as well to see a bit better how much effort it is to write new tests/features and back them up with step definitions. 

Have you uses Cucumber a lot? I am a bit skeptical towards this whole "executable specification" thing. I think [dhh](https://signalvnoise.com/posts/3159-testing-like-the-tsa) words almost come to my mind when I hear Cucumber:

> Don’t use Cucumber unless you live in the magic kingdom of non-programmers-writing-tests (and send me a bottle of fairy dust if you’re there!)

njam:
I don't have much experience with cucumber. I've written the same tests as these with vagrant-spec: https://github.com/vagrant-landrush/landrush/pull/145. See also https://github.com/vagrant-landrush/landrush/issues/136 for more information.
Any input welcome!

hferentschik:
> I don't have much experience with cucumber. I've written the same tests as these with vagrant-spec: #145. See also #136 for more information.

Interesting. I will have a look at both. Any idea why issue #136 is closed? Nothing seems to be merged yet in regards to acceptance tests. In fact I just created #180 to track and discuss this, but it seems we might be better off re-opening #136. WDYT?

For now I will still close this particular pull request, since we agree that in its current form we don't want to merge it.

njam:
agree!

njam:
Will be continued in: https://github.com/vagrant-landrush/landrush/pull/182

(can't reopen pull request if the first commit changed)

PR #146, opened 2016-01-01
Get changelog up to date

njam:
…also took the liberty to simplify the formatting a bit.

@ahpook @strzibny @fh @bexelbie please review

PR #147, opened 2016-01-01
Changing default TLD from vagrant.dev to vagrant.test

### LINKED ISSUE #118: Please use a different default domain than vagrant.dev
As `.dev` is one of the many new gTLDs introduced by ICANN it started to resolve to `127.0.53.53` (cf. #103) this domain should no longer be used by landrush.

See the following for further information:

* http://icannwiki.com/.dev
* https://www.icann.org/resources/pages/name-collision-2013-12-06-en#127.0.53.53 

A sensible alternative might be `.vagrant` or `.landrush`.

njam:
Based on #146
Closes #118

I suggest we should change the TLD, because ".dev" leads to problems, see #118.

@ahpook @strzibny @fh @bexelbie wdyt? please review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,26 +1,16 @@
 ## [Unreleased][unreleased]
-
-### Added
- - vagrant: support for libvirt provider (#138)
- - Add support for CNAME records (#99)
+- Added: Support for libvirt provider (#138)
+- Added: support for CNAME records (#99)
 
 ## [0.18.0] - 2015-01-24
-
-### Added
- - vagrant: support for `vagrant reload` (#101)
+- Added: support for `vagrant reload` (#101)
 
 ## [0.17.0] - 2015-01-18
-
-### Added
- - cli: `add` / `rm|del` subcommands (#96)
-
-### Fixed
- - cli: default to showing help when no command is specified
+- Added: cli `add` / `rm|del` subcommands (#96)
+- Fixed: cli: default to showing help when no command is specified
 
 ## [0.16.0] - 2015-01-18
-
-### Added
- - dns: support for IN::PTR records (#98)
+- Added: Support for IN::PTR records (#98)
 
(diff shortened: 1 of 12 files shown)


ahpook:
@njam Agree in principle - do you think this constitutes a semver breaking change though?

njam:
Hm, not sure. But since we're in "major version zero" we can do anything anyway ;)

> Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.
http://semver.org/

ahpook:
Success!

strzibny:
I actually think that the leading 0 version is not a good excuse since many projects use semver _with_ leading zero...

bexelbie:
As I understand anyone with an internet connected DNS resolver is broken with .dev, I believe this change is warranted.  However, I would agree with Shifting us to V1.0.0 to recognize the significance of the change.  I believe we should triage a few things to go with 1.0.0 if we go down that path.  I would nominate making landrush work on windows or officially noting it is broken and recommending a release with the old RubyDNS (or downgrading just for windows)

njam:
As you mentioned the current vagrant default TLD (.dev) is problematic for many users. I think we should release a new version with a different default soon.

For me both is okay, releasing as 0.19 or as 1.0!

Do you agree to use `.test`?

bexelbie:
@njam I feel like .local is a better choice than .test

ahpook:
> @njam <https://github.com/njam> I feel like .local is a better choice than .test
> 
> 


.local conflicts with the zerconf resolver on Mac OS X

bexelbie:
@ahpook Thanks for the follow up - after I posted that I found a conflict with a different daemon and then got pulled into meetings and forgot to update my comment :)

njam:
I had the suspicion that .local would conflict with *something*, thanks for clarifying!

So should we stick to `.test`? Personally I'm using a *real* TLD à la `my-product.dev.my-company.com`. But I think for a default `.test` is nice (and short).

bexelbie:
LGTM

PR #148, opened 2016-01-01
Add "minitest" dependency to gemfile

njam:
@ahpook @strzibny @fh @bexelbie please review

I think this is missing, as it's used in `test/test_helper.rb`.


njam:
@ahpook @strzibny @fh @bexelbie please review

fh:
@njam Sorry, missed this (the last weeks were rather busy here). lgtm!

Issue #16, opened 2013-11-11
windows: fork() function is unimplemented


isimmons:
First time using this and not sure if I'm even using it right :-)

Here is my vagrant file
```
# -*- mode: ruby -*-
# vi: set ft=ruby :

# Vagrantfile API/syntax version. Don't touch unless you know what you're doing!
VAGRANTFILE_API_VERSION = "2"

Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|

  config.vm.box = "precise32"
 
  config.vm.box_url = "http://files.vagrantup.com/precise32.box"

  config.landrush.enable

  config.vm.hostname = 'myhost.vagrant.dev'

  #config.vm.network :forwarded_port, guest: 80, host: 3000

  #config.vm.provision :shell, :path => "bootstrap.sh"
 
end
```
This is the output
```
  vagrant up
Bringing machine 'default' up with 'virtualbox' provider...
[default] Importing base box 'precise32'...
[default] Matching MAC address for NAT networking...
[default] Setting the name of the VM...
[default] Clearing any previously set forwarded ports...
[default] Creating shared folders metadata...
[default] Clearing any previously set network interfaces...
[default] [landrush] virtualbox requires an additional private network; adding it
[default] [landrush] starting dns server
[landrush] Starting daemon...
[default] Destroying VM and associated drives...
C:/Users/lotus/.vagrant.d/gems/gems/rexec-1.5.2/lib/rexec/daemon/controller.rb:75:in `fork': fork() function is unimplemented on this machine
 (NotImplementedError)
        from C:/Users/lotus/.vagrant.d/gems/gems/rexec-1.5.2/lib/rexec/daemon/controller.rb:75:in `start'
        from C:/Users/lotus/.vagrant.d/gems/gems/rexec-1.5.2/lib/rexec/daemon/base.rb:129:in `start'
        from C:/Users/lotus/.vagrant.d/gems/gems/landrush-0.7.0/lib/landrush/action/setup.rb:39:in `start_server'
        from C:/Users/lotus/.vagrant.d/gems/gems/landrush-0.7.0/lib/landrush/action/setup.rb:20:in `pre_boot_setup'
        from C:/Users/lotus/.vagrant.d/gems/gems/landrush-0.7.0/lib/landrush/action/setup.rb:8:in `block in call'
[...]

PR #154, opened 2016-01-15
Issue #16 - Getting Landrush to work on Windows

hferentschik:
This is a first stab at resolving #16 and getting Landrush to work on all OSes. The key changes are:

* Avoid the use of `fork` and use Ruby's `Process.spawn` to start the Landrush DNS server.  No dependency to `rexec` anymore
* Downgrading the `rubydns` gem to 0.8.5 which is based on `eventmachine` which works across OSes opposed to Celluloid which won't work on Windows. See also https://github.com/ioquatix/rubydns/#migrating-from-rubydns-08x-to-09x

Apart from that the changes of this pull request are consequences of the two points mentioned above. 

In `command.rb` I also removed the use of `/usr/bin/pr` (which does not exist on Windows) with `printf`. Last but not least, I added some documentation for configuring DNS on Windows. 

I tested against Windows 10 and it  works. I still need to test on Windows 7. There is no issue with running the Landrush DNS server, but rather a question of differences in the DNS configuration. On Windows 7 it might for example be necessary to configure the "DNS suffix list" for the adapter to which one adds the Landrush DNS server.

If we agree to merge this pull request, the next steps would be:
* Automate Windows DNS configuration
* Make tests runnable on Windows 
* Get greater test coverage
These things could be addressed in follow up issues.

Looking forward to get some feedback on this.


maxandersen:
Looks like the gemfile.lock picks up a newer RuboCop version that is not fully backwards compatible.

Seems adding something like what is outlined in this PR https://github.com/Homebrew/homebrew/pull/48144/files could work.

hferentschik:
> Looks like the gemfile.lock picks up a newer RuboCop version that is not fully backwards compatible.

TBH I did not run Rubocop. I think it was added just a while ago and I just rebased on top of these changes. Personally I only ever ran `rake` with explicit target which did not show any problems. However, looking at the `Rakefile` the `rubocop` task is part of the default task. 

You can run `rake build` or `rake test` directly to bypass this. I see what I can do regarding rubocop itself...

hferentschik:
I fixed some basic spacing issues, but other than that just re-generated `.rubocop_todo.yml`, since it looked like that most errors were just ignored anyways. Not sure whether all default Rubocop checks make sense anyways. Happy to discuss style once we have an overall agreement on this pull request.

hferentschik:
I am wondering what environment the CI job is running on. Linux? It reports:

    Finished tests in 96.890238s, 0.4335 tests/s, 0.6193 assertions/s.
    
      1) Failure:
    
    test_0004_stops the landrush server when there are no dependent machines left(    Landrush::Action::Teardown) [/home/travis/build/phinze/landrush/test/landrush/action    /teardown_test.rb:48]:
    
    Expected: false
    
      Actual: true
    
      2) Failure:
    
    test_0001_starts and stops a daemon(Landrush::Server::start/stop) [/home/travis/    build/phinze/landrush/test/landrush/server_test.rb:23]:
    
    Expected: false
    
      Actual: true
    
    42 tests, 60 assertions, 2 failures, 0 errors, 2 skips
    
    rake aborted!

My result of `rake test` on OS X is:

    Finished tests in 37.004713s, 1.1350 tests/s, 1.7025 assertions/s.

    42 tests, 63 assertions, 0 failures, 0 errors, 1 skips

I can try to run the tests in a Linux VM, but it would be nice to know the environment used for CI to make sure I align with it.

hferentschik:
Ok, I pushed another update to this pull request, fixing the tests on Linux as well. Turns out that I missed a `Process.detach` after spawning the Landrush DNS server. Seems OS X and Windows are more forgiving, where on Linux you end up with Zombie process which won't go away until the parent process exits.

hferentschik:
@strzibny - WDYT?

hferentschik:
Rebased on current master.

LalatenduMohanty:
@hferentschik works fine on Fedora 23. However I had to add gem 'json' to Gemfile 

```
[lmohanty@LalatenduM-laptop landrush-hardy]$ bundle exec rake
Running RuboCop...
Inspecting 44 files
............................................

44 files inspected, no offenses detected
Run options: --seed 7723

# Running:

...Stopping daemon...
Pid file /tmp/vagrant_landrush_test_working_dir/run/landrush.pid not found. Is the daemon running?
.Stopping daemon...
Pid file /tmp/vagrant_landrush_test_working_dir/run/landrush.pid not found. Is the daemon running?
.Stopping daemon...
...Stopping daemon...
Pid file /tmp/vagrant_landrush_test_working_dir/run/landrush.pid not found. Is the daemon running?
.Stopping daemon...
.Stopping daemon...
Pid file /tmp/vagrant_landrush_test_working_dir/run/landrush.pid not found. Is the daemon running?
.Stopping daemon...
.Stopping daemon...
..Stopping daemon...
.Stopping daemon...
.....Stopping daemon...
........S......Stopping daemon...
.Stopping daemon...
.Stopping daemon...
.SStopping daemon...
...

Finished in 16.901247s, 2.4850 runs/s, 3.5500 assertions/s.

42 runs, 60 assertions, 0 failures, 0 errors, 2 skips

You have skipped tests. Run with --verbose for details.
```

hferentschik:
> However I had to add gem 'json' to Gemfile to make it work

Hmm, interesting. I thought JSON is part of the standard lib, so it should not be necessary to install an additional gem. What Ruby version are you using?

LalatenduMohanty:
@hferentschik 
```
$ ruby --version
ruby 2.2.4p230 (2015-12-16 revision 53155) [x86_64-linux]
```

hferentschik:
Interesting. I guess it cannot harm to explicitly add 'json'. trying to understand why is it required. Did you specify any particular version?

LalatenduMohanty:
@hferentschik nope, it was just `gem install json` then added to Gemspec

LalatenduMohanty:
@hferentschik I think you need to update https://github.com/phinze/landrush/blob/master/CHANGELOG.md in this PR to list the fix

hferentschik:
Sure. The change log is not updated, nor the version bumped. This would 
happen
when a new release is cut. Probably making a bit of a version jump as 
well!?
Thanks for the feedback btw. Much appreciated.

hferentschik:
@LalatenduMohanty I've added the json gem to the Gemfile and pushed an update. Can you try once more?

strzibny:
You need json for Ruby packaged in Fedora, it's unbundled from Ruby itself.

hferentschik:
> You need json for Ruby packaged in Fedora, it's unbundled from Ruby itself.

Go figure.

LalatenduMohanty:
@hferentschik now `bundle exec rake` works fine in Fedora 23 without any modification

hferentschik:
> now bundle exec rake works fine in Fedora 23 without any modification 

nice

bexelbie:
@njam  @ahpook  @strzibny  @fh 
Please review - I am almost done with my review and really want to merge this so we can say we work on Windows again.

hferentschik:
Related to this issue is #157 which needs to be fixed as well.

.gitignore
```diff
@@ -1,9 +1,9 @@
-*.gem
+.ruby-gemset
+.ruby-version
 *.rbc
 .bundle
 .config
 .yardoc
-Gemfile.lock
```
fh: @hferentschik - Why did you add the Lockfile here?

fh:
@hferentschik Is there a reason why you added the style violations to the rubocop todo? The whole point behind rubocop is to enforce a consistent style across the new code - the todo is only there for old code that has to be cleaned and refactored, so I would prefer not to add more debt to that.

.gitignore
hferentschik: I removed it from .gitignore and added it to the repo. IMO in most cases the Gemfile.lock should be checked in. Only this makes builds repeatable. There is an argument to be made whether the lock file should not be checked in got gem development, but I prefer to lock the dependencies down ensuring anyone who wants to build the source can do so with the right dependencies. Verifying potential dependency changes can then be done by a dedicated person or at dedicated times where one also can then address potential incompatibilities. See also https://viget.com/extend/bundler-best-practices

Bottom line, I think `Gemfile.lock` belongs under version control. This is, however, independent of this issue for course. If the majority of people is against it, I am fine changing this. Hence I added it as last commit. I could also imagine creating a dedicated issue and pull request to separate this out.

hferentschik:
> Is there a reason why you added the style violations to the rubocop todo? 

Yes and no. The rubocop todo file needed upgrading anyways, due to some changes in its format. Given the amount of violation which were already in there, I felt it was not worth spending too much time on changed violations. I am also very much for a code style, but not necessarily agree with all the rubocop defaults. 

That said, if the plan is to stick to all rubocop defaults, I am happy to go over my changes again and adjust the code to at least come to the same level (if not better) of violations as prior the pull request. I just did not want to spend time on style problems until I get some feedback on the actual technical changes. WDYT?

fh:
@hferentschik Sounds fair - maybe just use the autocorrect feature to remove the overall amount of errors, and we'll see for the rest. Sorry for giving style comments first, I am just not able to give the PR a test drive for the next days - as soon as I am able to do so, there will be more substantial feedback (unless someone else already chimed in ;))

Fwiw, I dont like all the rubocop defaults either, and I dont have any preference in regards to which to use - I just like to have something, and improving from there.

hferentschik:
> maybe just use the autocorrect feature to remove the overall amount of errors

+1 

> Sorry for giving style comments first, I am just not able to give the PR a test drive for the next days

Sure. No problem. Just happy to get some feedback. I also found some other related issues which I will address in follow up pull requests. For example the way the Vagrant data dir is determined is wrong. One need to go via Vagrant `@env` instance which is passed to the plugin. This does not really effect this code, but can lead to problems on Windows where you want to start Landrush from a cygwin shell.  

> I just like to have something, and improving from there. 

+1 Sounds like a good plan.

lib/landrush/server.rb
```diff
@@ -30,38 +55,85 @@ def self.upstream
...
     end
 
-    def self.pid
-      RExec::Daemon::ProcessFile.recall(self)
+    # Used to start the Landrush DNS server as a child process using ChildProcess gem
+    def self.start
+      ensure_path_exits(log_file)
+
+      if OS.windows?
+        pid = spawn('ruby', "#{__FILE__}", "#{port}", "#{working_dir}", :chdir => working_dir.to_path, [:out, :err] => [log_file, "w"], :new_pgroup => true)
```
strzibny: Can we use the new Hash syntax here?

lib/landrush/server.rb
```diff
@@ -30,38 +55,85 @@ def self.upstream
...
+      else
+        pid = spawn('ruby', "#{__FILE__}", "#{port}", "#{working_dir}", :chdir => working_dir.to_path, [:out, :err] => [log_file, "w"], :pgroup => true)
```
strzibny: Why writing `"#{port}"` instead of just `port`? Same for `working_dir`...

lib/landrush/server.rb
```diff
@@ -83,9 +155,90 @@ def self.run
...
+      else
+        transaction.respond!(Name.create(value), resource_class: IN::CNAME, ttl: 0)
+        check_a_record(value, transaction)
+      end
+    end
+
+    # private methods
+    def self.write_pid(pid)
+      ensure_path_exits(pid_file)
+      File.open(pid_file, 'w') {|f| f << pid.to_s}
```
strzibny: Another smallish style thing: `{ |f| f << pid.to_s }` instead of `{|f| f << pid.to_s}`

strzibny:
I included few styles issues that I would like to change. As for real Windows testing, I don't have the Windows machine for testing that I had. @pvalena can you please test this patch?

pvalena:
@strzibny Hi, sorry i am really busy right now, but as soon as I get it done I will test it.

hferentschik:
@strzibny, thanks for the feedback. I am planning another update to this pull request for next week. I'll address your style comment and try to reduce the overall amount of robocup violations.

.gitignore
njam: Is vagrant using the lockfile when resolving the dependencies? I guess having it in the repo only makes sense if it does.

hferentschik:
@fh, @strzibny - I did a fair bit of code cleanup and reduced the amount of rubocop violations considerably. Forced pushed an update to this pull request.

Also created a follow up pull request #160 based on this pull request addressing issue #157.

@bexelbie How is the review going?

hferentschik:
Here is another idea. The problem seems to be access for people to Windows machines. What's about marking Windows support as experimental? If we can agree that the code is not breaking anything for Linux or OS X, we could then go ahead and merge. We could even do a release and this way get feedback from people trying to use Landrush on Windows.

bexelbie:
@hferentschik moving to windows as experimental could work - I am traveling this week so my windows testing time is more limited :(

strzibny:
@hferentschik like that idea very much.

hferentschik:
> like that idea very much.

Ok, let's go for something like that then. I guess all that's needed are some notes in the docs

bexelbie:
@hferentschik do you have time to rebase and add the docs notes?  I saw your comment about wanting to land this, this week.

hferentschik:
> do you have time to rebase and add the docs notes?

will do. I also will add a commit which allows to run the tests on Windows. Now you can develop Landrush on Windows as well :-)

I also confirmed the setup once more on a "virgin" Windows 10 setup. Worked fine.

> I saw your comment about wanting to land this, this week.

Most likely tomorrow. Once everything is merged we will need to talk about a release :-)

hferentschik:
Ok, I pushed one more update adding a note to the docs that Windows support is experimental for now. I also changed the test suite and it runs now on Windows as well. So Landrush can now be run and developed on Windows. 

I would like to merge this now. Can I get one more maintainer ACK - @njam, @ahpook, @strzibny, @fh, @bexelbie, @phinze?

ahpook:
@hferentschik looks like there's merge conflicts, according to github - do you need to rebase? I'm 👍 once that's resolved, I read through the docs and all seems reasonable.

hferentschik:
>  looks like there's merge conflicts, according to github - do you need to rebase? 

I don't think so. I might have pushed one time too many to this branch. I'll can close this pull request and open a new one. That should sort things out.

> I'm :+1: once that's resolved, I read through the docs and all seems reasonable.

Thanks.

hferentschik:
Actually I needed one more rebase. Nevertheless, here is the new pull request, passing also Travis CI - https://github.com/vagrant-landrush/landrush/pull/170

PR #155, opened 2016-01-23
Upgrade vagrant dev dependency to 1.8.1

njam:


njam:
@ahpook @strzibny @fh @bexelbie please review

strzibny:
I have no problem with moving this forward although I personally test with version packaged in Fedora which is older (but also older then current dev version). Specifying RuboCop version is also +1 from me.

njam:
I'm also still running 1.7 due to some other plugins being incompatible with 1.8.
But if you run `bundle exec vagrant <args>` in the landrush directory you can also test it with the version mentioned in the `Gemfile`.

Issue #157, opened 2016-02-11
Landrush uses potentially wrong data directory


hferentschik:
Landrush uses the following to determine the data dir:

    File.expand_path('~/.vagrant.d/data/landrush')

This won't work, if the user uses environment variables to change the location of this `.vagrant.d`directory. It is also problematic on Windows when using cgwin in which case the cygwin home directory is used, instead of the Windows home directory (C:/Home/<user>).

The plugin should make of the `@env` instance giving access to `Vagrant::Environment`.

PR #172, opened 2016-04-12
Issue #157 Making sure that the right Vagrant data dir is chosen 

hferentschik:
Refactored the tests to set the required directories during test setup resp delete them during tear down.


hferentschik:
Here is the updated pull request for issue #157. Basically it avoids the use of `~/.vagrant.d` to determine the location of the user's _vagrant.d_ directory. This causes problems on Windows where the _~_ resolves to different directories depending on whether Cygwin or PowerShell is used. The problem can even occur on OS X or Linux if _VAGRANT_HOME_ is used to set a custom _vagrant.d_ location. One needs to use the values passed by the Vagrant environment, `env[:home_path]` resp. `@env.data_dir` (for some reason `@env.data_dir` is not always available).

@njam, @ahpook, @strzibny, @fh, @bexelbie, @phinze - thoughts?

bexelbie:
I like this.  It would be nice to land a small bit of docs around how to fix this error when raised:

https://github.com/vagrant-landrush/landrush/pull/172/files#diff-695e5af033f2e918c112401dcbf9bfb1R15

hferentschik:
>  It would be nice to land a small bit of docs around how to fix this error when raised

Ok. From actual user point of view it cannot happen really. The user interacts via the various plugin commands with the Server. He never starts one directly. It is more likely to occur in test scenarios where single components get tested. In this case one needs to make sure the working directory gets set to a tmp directory. I can add some pointers in code and the development section of the README. @bexelbie would that work for you?

bexelbie:
@hferentschik yes

hferentschik:
@bexelbie I added some better comments into the code and also created a follow up issue https://github.com/vagrant-landrush/landrush/issues/178

bexelbie:
@hferentschik thanks for merging - i was traveling and this fell off my radar.

PR #182, opened 2016-04-24
Acceptance tests: cucumber

njam:
Continuation of #144, part of #136

Run with:
```
bundle exec rake features
```

### DIFF
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -1,0 +1,15 @@
+require 'aruba/cucumber'
+require 'komenda'
+
+Aruba.configure do |config|
+  config.exit_timeout = 300
+  config.activate_announcer_on_command_failure = [:stdout, :stderr]
+end
+
+After do |_scenario|
+  Komenda.run('bundle exec vagrant landrush stop', fail_on_fail: true)
+
+  if File.exist?(File.join(aruba.config.working_directory, 'Vagrantfile'))
+    Komenda.run('bundle exec vagrant destroy -f', cwd: aruba.config.working_directory, fail_on_fail: true)
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -24,4 +24,7 @@ group :development do
   gem 'byebug'
   gem 'mocha'
   gem 'minitest'
+  gem 'cucumber', '~> 2.1'
+  gem 'aruba', '~> 0.13'
+  gem 'komenda', '~> 0.1.6'
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -30,17 +30,39 @@ PATH
 GEM
   remote: https://rubygems.org/
   specs:
+    aruba (0.14.1)
+      childprocess (~> 0.5.6)
+      contracts (~> 0.9)
+      cucumber (>= 1.3.19)
+      ffi (~> 1.9.10)
+      rspec-expectations (>= 2.99)
+      thor (~> 0.19)
     ast (2.2.0)
     builder (3.2.2)
     byebug (8.2.1)
     childprocess (0.5.9)
       ffi (~> 1.0, >= 1.0.11)
+    contracts (0.13.0)
+    cucumber (2.3.3)
+      builder (>= 2.1.2)
+      cucumber-core (~> 1.4.0)
+      cucumber-wire (~> 0.0.1)
+      diff-lcs (>= 1.1.3)
+      gherkin (~> 3.2.0)
+      multi_json (>= 1.7.5, < 2.0)
+      multi_test (>= 0.1.2)
+    cucumber-core (1.4.0)
+      gherkin (~> 3.2.0)
+    cucumber-wire (0.0.1)
+    diff-lcs (1.2.5)
     domain_name (0.5.25)
       unf (>= 0.0.5, < 1.0.0)
     erubis (2.7.0)
     eventmachine (1.0.8)
+    events (0.9.8)
     ffi (1.9.10)
     ffi (1.9.10-x64-mingw32)
+    gherkin (3.2.0)
     gssapi (1.2.0)
       ffi (>= 1.0.1)
     gyoku (1.3.1)
@@ -51,6 +73,8 @@ GEM
     httpclient (2.7.1)
     i18n (0.7.0)
     json (1.8.3)
+    komenda (0.1.6)
+      events (~> 0.9.8)
     listen (3.0.5)
       rb-fsevent (>= 0.9.3)
       rb-inotify (>= 0.9)
@@ -66,6 +90,7 @@ GEM
     mocha (1.1.0)
       metaclass (~> 0.0.1)
     multi_json (1.11.2)
+    multi_test (0.1.2)
     net-scp (1.1.2)
       net-ssh (>= 2.6.5)
     net-sftp (2.1.2)
@@ -96,6 +121,10 @@ GEM
       http-cookie (>= 1.0.2, < 2.0)
       mime-types (>= 1.16, < 3.0)
       netrc (~> 0.7)
+    rspec-expectations (3.4.0)
+      diff-lcs (>= 1.2.0, < 2.0)
+      rspec-support (~> 3.4.0)
+    rspec-support (3.4.1)
     rubocop (0.38.0)
       parser (>= 2.3.0.6, < 3.0)
       powerpack (~> 0.1)
@@ -107,6 +136,7 @@ GEM
       eventmachine (~> 1.0.0)
     rubyntlm (0.4.0)
     rubyzip (1.1.7)
+    thor (0.19.1)
     unf (0.1.4)
       unf_ext
     unf_ext (0.0.7.1)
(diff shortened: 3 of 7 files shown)


njam:
I've rewritten the cucumber step definitions to check DNS lookups to be based on Ruby libraries. The check to query against Landrush uses `Resolv::DNS` and the check to query on the operating system level uses `Addrinfo`.

I think this way it should be cross-platform compatible. On OSX the tests run fine, could you test it on Windows? Note that the example with `ubuntu/wily64` fails because of a bug in Landrush.

@hferentschik @fh wdyt?

hferentschik:
> I've rewritten the cucumber step definitions to check DNS lookups to be based on Ruby libraries. The check to query against Landrush uses Resolv::DNS and the check to query on the operating system level uses Addrinfo.

Nice. I'll have a look tomorrow.

hferentschik:
@njam, I tried these changes locally and it might be that the Cucumber approach is maybe even better than vagrant-spec. A couple of things though.

When trying to run the tests for the first time on my OS X machine, I got:

     ubuntu/wily64   | virtualbox |
          expected "bundle exec vagrant up --provider virtualbox" to be successfully     executed (RSpec::Expectations::ExpectationNotMetError)
          features/dns_resolution.feature:24:in `When I successfully run `bundle exec     vagrant up --provider virtualbox`'
          features/dns_resolution.feature:16:in `When I successfully run `bundle exec     vagrant up --provider <provider>`'
    
    Failing Scenarios:
    cucumber features/dns_resolution.feature:23 # Scenario Outline: booting a box,     Examples (#1)
    cucumber features/dns_resolution.feature:24 # Scenario Outline: booting a box,     Examples (#2)
    
    2 scenarios (2 failed)
    10 steps (2 failed, 6 skipped, 2 passed)
    0m44.665s

The VM would not even comes up. The reason is that _ResolverConfig_ tries to add an entry to _/etc/resolver/_ and needs a password for executing in sudo mode. After the first time it works. I think there are a couple of things at play here. First the sudo password is cached, but also _/etc/resolver/my-tld_ never gets cleaned up. Would there be a way handle the sudo part somehow via the test? Or maybe it would be better to change the code in a way that it does not even create anything in _/etc/resolver_ under test. WDYT?

Once the VM actually starts, I am seeing:

        Examples: 
          | debian/jessie64 | virtualbox |
          | ubuntu/wily64   | virtualbox |
          
          expected: "10.10.10.123"
               got: "10.0.3.1"
          
          (compared using ==)
           (RSpec::Expectations::ExpectationNotMetError)
          ./features/step_definitions/dns.rb:7:in `/^the hostname "([^"]+)" should     resolve to "([^"]+)" on the internal DNS server$/'
          features/dns_resolution.feature:24:in `Then the hostname "my-host.my-tld"     should resolve to "10.10.10.123" on the internal DNS server'
          features/dns_resolution.feature:17:in `Then the hostname "my-host.my-tld"     should resolve to "10.10.10.123" on the internal DNS server'
    
    Failing Scenarios:
    cucumber features/dns_resolution.feature:24 # Scenario Outline: booting a box,     Examples (#2)
    
    2 scenarios (1 failed, 1 passed)
    10 steps (1 failed, 2 skipped, 7 passed)
    1m59.587s

Is this the bug you are referring to? Any idea why _10.0.3.1_ is returned? Is there an issue for that?

features/support/env.rb
```diff
@@ -0,0 +1,15 @@
+require 'aruba/cucumber'
+require 'komenda'
+
+Aruba.configure do |config|
+  config.exit_timeout = 300
+  config.activate_announcer_on_command_failure = [:stdout, :stderr]
+end
+
+After do |_scenario|
+  Komenda.run('bundle exec vagrant landrush stop', fail_on_fail: true)
```
hferentschik: What is Komenda about in this case? What does it add to the mix?

hferentschik:
@njam are you running Linux or OS X?

features/support/env.rb
njam: It's just a simple way to run a shell command. I couldn't easily find something exposed by Aruba (which might be nicer?).

njam:
> are you running Linux or OS X?

OSX

> Or maybe it would be better to change the code in a way that it does not even create anything in /etc/resolver under test. WDYT?

I was thinking about this as well. It feels a bit strange that we're relying on global state that is persisted on the operating system for these tests. But then on the other hand these are acceptance tests, and their purpose is to verify that Landrush works correctly, in combination with the operating system. I think it's the only way to let Landrush do what it usually does if we want to determine if that works or not.

> needs a password for executing in sudo mode. 

Hmmm I see. Seems like the way aruba executes the process as a ["leader"](https://github.com/cucumber/aruba/blob/master/lib/aruba/processes/spawn_process.rb#L72) which means it doesn't have a tty?
I think it would be best if we would ask for sudo, because we don't know what happens within Landrush for the current platform (linux, osx, windows). Should we look into ways to make Aruba commands have a TTY?

> Is this the bug you are referring to? Any idea why 10.0.3.1 is returned? Is there an issue for that?

Yes, IP detection on the ubuntu machine doesn't work correctly. It picks up the "wrong" interface's address. There's no specific ticket about this configuration. But I guess we're picking the wrong interface in a lot of configurations. I was referring to https://github.com/vagrant-landrush/landrush/pull/122 as a potential solution.

hferentschik:
> I was thinking about this as well. It feels a bit strange that we're relying on global state that is persisted on the operating system for these tests. But then on the other hand these are acceptance tests, and their purpose is to verify that Landrush works correctly,

Sure, its a tricky one. But if we really also modify the state of the host, we also need to make sure that the changes are reset after the test. BTW, if the aim is to test the host setup as well, then the test (at least for the OS X case) is not quite right. The DNS resolution is against the Landrush DSN server port (10053), not the default DNS port. On OS X one would need to check against the default DNS port to verify whether the DNS configuration change via _/etc/resolver_ worked. 

For Linux and Windows the host configuration is not yet automated, so there one needs to verify against port 10053. I am working on some automation for the host configuration on Windows, but that is still a bit off. 

Personally, I think it might be easier for now to offer a way to skip host configuration updates (add something like _config.landrush.skip.host.dns.config_) and test "just" that the Landrush DNS server works.

This removes the need to proper test cleanup (which will differ by OS) and it also works around the sudo permission problem.

We can then evolve the tests over time.

> Yes, IP detection on the ubuntu machine doesn't work correctly. It picks up the "wrong" interface's address.

Ok. I see.

BTW, I will be traveling next week, so I won't be so responsive, but it would be great if we could get something worked out for the week after next, so that we can move on with cutting a release. WDYT?

njam:
> if the aim is to test the host setup as well, then the test (at least for the OS X case) is not quite right.

I think it is correct actually, as it is testing both cases:
- *the hostname "([^"]+)" should resolve to "([^"]+)" on the internal DNS server*: Checks that Landrush's DNS server resolves to the desired address
- *the hostname "([^"]+)" should resolve to "([^"]+)" on the host*: Checks that Ruby's standard way of resolving an address has the same result. On OSX this will be using mDNS and `/etc/resolver`.

`config.landrush.skip.host.dns.config` sounds like an idea. I'm just a bit concerned that we will be skipping a lot of important functionality that would be great to cover with an acceptance test.

I would rather suggest to *keep* the check about the resolution on the default resolver in the test. These tests are not running in CI, but are rather meant for us developers, to have a tool to verify that everything works. So one could argue that *if* somebody is running those tests on Windows, they should also set up DNS resolution. And if they don't they will understand why the tests fail and they can ignore that if they want to.

wdyt?

hferentschik:
> I think it is correct actually, as it is testing both cases:

You are right. Sorry. So the problem is then actually the other way around. On Linux and Windows these tests will fail, because there the host visibility is not setup automatically.

> I would rather suggest to keep the check about the resolution on the default resolver in the test. These tests are not running in CI, but are rather meant for us developers, to have a tool to verify that everything works. So one could argue that if somebody is running those tests on Windows, they should also set up DNS resolution. And if they don't they will understand why the tests fail and they can ignore that if they want to.

I see your point. As you say, for now these tests will only be for development and it is fair to assume that the  developer has to set up host visibility manually for the OSes where it is not supported out of the box.

That we cannot run these tests on CI is a shame, but that would require that we find a CI provider which either offers bare metal CI servers or setups where one can run virtualization within virtualization. Travis-CI at least does not allow to run VirtualBox on top of their CI environment.

@njam, do you think we can at least address the sudo issue? Also, should we really add 'ubuntu/wily64' as an example even though we know it fails? I think we should also expand the README in regards to who should use the cucumber tests and how to use them (aka the need to configure DNS on the host for Linux and Windows). WDYT?

hferentschik:
I went ahead and merged the pull request with some minor modifications. It works on OS X (with the potential sudo issue) and Linux. On the latter one needs to manually do the dnsmasq configuration. On Windows it won't work at all atm :-( due to a bug in Aruba - https://github.com/cucumber/aruba/issues/387

vagrant-spec actually would work on Windows (apart of the fact that one would need to also adjust all the Unix type commands in the vagrant-spec based pull request), however, I think the syntax and setup is better with Cucumber and Aruba. Also the Aruba community seems to be active and interested in resolving the problem. So hopefully the test will work in the near future. 

@njam, thanks for the pull request.

njam:
👍 Thanks! 

> Travis-CI at least does not allow to run VirtualBox on top of their CI environment.

Agree, I don't know of any free service that would allow for this. We would probably need to spend some money and set up a Jenkins which does that for us.

Issue #165, opened 2016-03-04
Unable to use with CentOS v5 boxes on vmware_fusion


khaije1:
I get a similar error with any CentOSv5 VM I've tried.
The below error happens while using this public CentOSv5 VM, but my packer-generated VM experiences the exact same problem.

With minor testing it appears that the 'ip' variable is being passed to '_gateway_for_ip(ip)' as a nil. 

```
mentat:genesis-ctl khaije1$ vagrant init bento/centos-5.11 ; vagrant up
Bringing machine 'ness-scan-p1' up with 'vmware_fusion' provider...
==> ness-scan-p1: Cloning VMware VM: 'bento/centos-5.11'. This can take some time...
==> ness-scan-p1: Checking if box 'bento/centos-5.11' is up to date...
==> ness-scan-p1: Verifying vmnet devices are healthy...
==> ness-scan-p1: [landrush] Host DNS resolver config looks good.
==> ness-scan-p1: [landrush] starting dns server
[landrush] Starting daemon...
[landrush] Waiting for daemon to start...
==> ness-scan-p1: Preparing network adapters...
==> ness-scan-p1: Starting the VMware VM...
==> ness-scan-p1: Waiting for machine to boot. This may take a few minutes...
    ness-scan-p1: SSH address: 172.16.105.129:22
    ness-scan-p1: SSH username: vagrant
    ness-scan-p1: SSH auth method: private key
    ness-scan-p1: Warning: Connection refused. Retrying...
    ness-scan-p1: 
    ness-scan-p1: Vagrant insecure key detected. Vagrant will automatically replace
    ness-scan-p1: this with a newly generated keypair for better security.
    ness-scan-p1: 
    ness-scan-p1: Inserting generated public key within guest...
    ness-scan-p1: Removing insecure key from the guest if it's present...
    ness-scan-p1: Key inserted! Disconnecting and reconnecting using new SSH key...
==> ness-scan-p1: Machine booted and ready!
==> ness-scan-p1: [landrush] setting up machine's DNS to point to our server
==> ness-scan-p1: Stopping the VMware VM...
==> ness-scan-p1: Deleting the VM...
/Users/khaije1/.vagrant.d/gems/gems/landrush-0.18.0/lib/landrush/action/redirect_dns.rb:34:in `_gateway_for_ip': undefined method `split' for nil:NilClass (NoMethodError)
[...]

PR #191, opened 2016-05-31
Issue #165 Support for CentOS v5 boxes 

khaije1:
Minor changes to enable Linux CentOS v5 to work correctly with Landrush.

Tested OK on CentOS versions 5, 6, 7

### DIFF
--- a/lib/landrush/cap/linux/add_iptables_rule.rb
+++ b/lib/landrush/cap/linux/add_iptables_rule.rb
@@ -3,7 +3,7 @@ module Landrush
     module Linux
       module AddIptablesRule
         def self.add_iptables_rule(machine, rule)
-          _run(machine, %Q(iptables -C #{rule} 2> /dev/null || iptables -A #{rule}))
+          _run(machine, %Q(/sbin/iptables -C #{rule} || /sbin/iptables -A #{rule}))
         end
 
         def self._run(machine, command)
--- a/lib/landrush/cap/linux/configured_dns_servers.rb
+++ b/lib/landrush/cap/linux/configured_dns_servers.rb
@@ -4,9 +4,9 @@ module Landrush
       module ConfiguredDnsServers
         def self.configured_dns_servers(machine)
           return @dns_servers if @dns_servers
-          machine.communicate.sudo('cat /etc/resolv.conf | grep ^nameserver') do |type, data|
+          machine.communicate.sudo('sed -ne \'s/^nameserver \([0-9.]*\)$/\1/p\' /etc/resolv.conf') do |type, data|
             if type == :stdout
-              @dns_servers = Array(data.scan(/\d+\.\d+\.\d+\.\d+/))
+              @dns_servers = Array(data.scan(/\d+\.\d+\.\d+\.\d+/)) unless data.to_s.empty?
             end
           end
           @dns_servers
--- a/lib/landrush/cap/linux/read_host_visible_ip_address.rb
+++ b/lib/landrush/cap/linux/read_host_visible_ip_address.rb
@@ -39,7 +39,7 @@ module Landrush
         end
 
         def self.command
-          %Q(hostname -I)
+          %Q(hostname -I 2>/dev/null || /sbin/ip ad | sed -ne 's/^\\W*inet \\([0-9.]*\\).* global .*$/\\1/p' 2>/dev/null)
         end
       end
     end


hferentschik:
thanks @khaije1 for the pull request. Looking into it asap.

> Tested OK on CentOS versions 5, 6, 7

How does this effect other non CentOS Linux versions?

khaije1:
All the tools and paths are generally stable across distros. 
In particular, the `ip` utility has been around for a long long time and it's presence should be a safe assumptions. I'm not thrilled about explicitly declaring the path to iptables, (or even relying on it where firewalld is of some prominence), but it's doesn't do anything to worsen or diminish the way these situation are handled.

I'll plan to test on a variety of debian and fedora -based distros. Any particular suggestions?

lib/landrush/cap/linux/read_host_visible_ip_address.rb
```diff
@@ -39,7 +39,7 @@ def self.read_host_visible_ip_address(machine)
         end
 
         def self.command
-          %Q(hostname -I)
+          %Q(hostname -I 2>/dev/null || /sbin/ip ad | sed -ne 's/^\\W*inet \\([0-9.]*\\).* global .*$/\\1/p' 2>/dev/null)
```
hferentschik: You cannot really assume _ip_ is installed. This was a long standing discussion. It is not. The whole of _read_host_visible_ip_address_ will be changed by pull request #193. We will be introducing a new guest capability which will copy a binary onto the guest which will determine the available interfaces.

lib/landrush/cap/linux/configured_dns_servers.rb
```diff
@@ -4,9 +4,9 @@ module Linux
       module ConfiguredDnsServers
         def self.configured_dns_servers(machine)
           return @dns_servers if @dns_servers
-          machine.communicate.sudo('cat /etc/resolv.conf | grep ^nameserver') do |type, data|
+          machine.communicate.sudo('sed -ne \'s/^nameserver \([0-9.]*\)$/\1/p\' /etc/resolv.conf') do |type, data|
```
hferentschik: So why this change. What is the actual issue you are addressing? The _cat_ command did return something wrong on CentOS? I

lib/landrush/cap/linux/add_iptables_rule.rb
```diff
@@ -3,7 +3,7 @@ module Cap
     module Linux
       module AddIptablesRule
         def self.add_iptables_rule(machine, rule)
-          _run(machine, %Q(iptables -C #{rule} 2> /dev/null || iptables -A #{rule}))
+          _run(machine, %Q(/sbin/iptables -C #{rule} || /sbin/iptables -A #{rule}))
```
hferentschik: So why the full path?

hferentschik:
> I'll plan to test on a the oldest supported and the most recent versions of debian, arch, coreos, ubuntu, and fedora. Any particular suggestions?

All :-) But seriously, using the cucumber test framework we already have - https://github.com/vagrant-landrush/landrush/blob/master/features/dns_resolution.feature

If you keep adding boxes to the [example list](https://github.com/vagrant-landrush/landrush/blob/master/features/dns_resolution.feature#L23) you can verify the DNS setup for all machine types you like. This will of course take some time and it is probably not something we want to have enabled all the time, but one can add them to see how it goes and then comment everything out except the most popular ones.

lib/landrush/cap/linux/read_host_visible_ip_address.rb
khaije1: This PR can be considered complimentary to the binary solution proposed in #193.

The `ip` utility was mentioned [here](https://github.com/vagrant-landrush/landrush/pull/193#issuecomment-225037763), but I dont see any information designating it as an unsuitable for this purpose. According to the man page it was added alongside linux 2.2 in 1999.  On Linux, `ip` is the long-term replacement for `ifconfig`, so is a better option in all respects.

lib/landrush/cap/linux/add_iptables_rule.rb
khaije1: The default path on CentOS v5 is "/usr/kerberos/bin:/usr/local/bin:/bin:/usr/bin".

lib/landrush/cap/linux/configured_dns_servers.rb
khaije1: This change is basically unimportant.

hferentschik: ok

lib/landrush/cap/linux/read_host_visible_ip_address.rb
hferentschik: I merged the landrush-ip integration now. This change is obsolete.

hferentschik:
@khaije1, I merged pull request #193. Would you mind verifying latest master with CentOS guests? Let's check what changes are still required (if any).

khaije1:
I'm stuck behind https://github.com/mitchellh/vagrant/issues/6775, but will test at next opportunity.

khaije1:
Will test/verify at next opportunity.

hferentschik:
@khaije1 any update on this?

khaije1:
Unfortunately I can only report that the latest v1 release, it seems #193 is in https://github.com/vagrant-landrush/landrush/tree/v1.0.0 ... is that correct? In my recent testing v1.0.0 isn't working on CentOS version 5, 6, or 7. I'd like to dig further into why but haven't had time to. 

The VM's will now progress as normal but once booted, DNS resolution is broken for landrush-contrived A records and for public DNS.

I've switched to using #191 which is working well with vagrant v1.8.5 and vagrant-vmware-fusion 4.0.10, in order to be able to continue with my day-to-day work.

hferentschik:
Closing this pull request, since it needs more work. Happy to look at an updated pull request.

Issue #189, opened 2016-05-19
Improve IP determination by using config information


hferentschik:
In the case of a statically defined IP, we should be able to extract the IP from the Landrush config.

hferentschik:
Merged

PR #192, opened 2016-06-02
Issue #189 Use static private network configuration to find routable host IP

hferentschik:
In the case of a private network with static IP, it is possible to extract the routable host IP from Vagrant's network config. Only if there is no static network config or the configuration is not complete (eg multiple private networks where none is explicitly selected via _config.landrush.host_ip_address_ we move on and rely on the read_host_visible_ip_address capability. Currently this is still just based on _uname_,but will be hopefully shortly extended by @Werelds IP selection approach - see pull request #122

### DIFF
--- a/lib/landrush/action/setup.rb
+++ b/lib/landrush/action/setup.rb
@@ -54,7 +54,7 @@ module Landrush
 
       def setup_static_dns
         config.hosts.each do |hostname, dns_value|
-          dns_value ||= machine.guest.capability(:read_host_visible_ip_address)
+          dns_value ||= host_ip_address
           unless Store.hosts.has?(hostname, dns_value)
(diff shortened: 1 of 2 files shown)


hferentschik:
@bexelbie WDYT?

bexelbie:
If I am understanding #122, it will return this information.  Why have a separate detection routine just for this one specific scenario?

hferentschik:
> Why have a separate detection routine just for this one specific scenario?

Because for the case where the private network is defined statically, there is no need to go via the guest capabilities. This route seems safer to me. Once DHCP comes into play you need to contact the guest. 

Also if the user defines multiple private network you might get into trouble with the NIC detection approach. Much easier to stick to the existing config and an already existing config option.

Issue #171, opened 2016-04-11
Provide automatic host visibility configuration on Windows


hferentschik:
Similar to OS X (where _/etc/resolver_), the plugin should automatically configure the host's DNS setup to make Landrush visible on the Windows host.

PR #204, opened 2016-06-28
Issue #171 Adding support for automatic configuration of host visibility on Windows

hferentschik:
- Making use of netsh and Windows registry to modify network configuration
- Adding test harness

This works on Win 10. Needs review and if possible more testing.


hferentschik:
@budhrg @bexelbie any comment? Probably hard to tell, since this is very Windows specific code, but maybe you have some general feedback or the ability to test on Windows.

budhrg:
@hferentschik Could you refer issue in the description? Handy in clicking instead of going through.

Gemfile.lock
```diff
@@ -1,175 +0,0 @@
-GIT
```
budhrg: Why `Gemfile.lock` removed?

test/landrush/win_network_config_test.rb
```diff
@@ -0,0 +1,76 @@
+require_relative '../test_helper'
+
+module Landrush
+  TEST_IP = '10.42.42.42'.freeze
+
+  describe WinNetworkConfig do
+    before do
+      @vboxmanage_found = !Landrush::WinNetworkConfig.which('VBoxManage').nil?
+      @admin = Landrush::WinNetworkConfig.admin_mode?
```
budhrg: something like `@admin_enabled`

lib/landrush/win_network_config.rb
```diff
@@ -0,0 +1,159 @@
...
+      else
+        @env[:ui].info("[landrush] #{msg}")
+      end
+    end
+  end
+end
+
+# Only run the following code when this file is the main file being run
+# instead of having been required or loaded by another file
+if __FILE__ == $0
```
budhrg: Is this the trick you were talking about>
What is its purpose?

lib/landrush/win_network_config.rb
```diff
@@ -0,0 +1,159 @@
...
+      end
+      return nil if network_details[0].nil?
+      network_details[0].split(/\n/)[0].match(/Configuration for interface "(.*)"/).captures[0].strip
+    end
+
+    # Makes sure that we have admin privileges and if nor starts a new shell with the required
+    # privileges
+    def ensure_admin_privileges(file, *args)
+      unless self.class.admin_mode?
+        require 'win32ole'
```
budhrg: See you did that `require` :smile:

lib/landrush/win_network_config.rb
```diff
@@ -0,0 +1,159 @@
...
+      network_guid = get_guid(network_name)
+      interface_path = INTERFACES + "\\{#{network_guid}}"
+      Win32::Registry::HKEY_LOCAL_MACHINE.open(interface_path, Win32::Registry::KEY_ALL_ACCESS) do |reg|
+        reg['NameServer'] = name_server
+        reg['Domain'] = domain
+      end
+    end
+
+    # Cross-platform way of finding an executable in the $PATH.
+    def self.which(cmd)
```
budhrg: You may want to use vagrant supported `Which` module https://github.com/mitchellh/vagrant/blob/master/lib/vagrant/util/which.rb

Somehow it uses similar code as you are using :smile:

lib/landrush/win_network_config.rb
```diff
@@ -0,0 +1,159 @@
...
+        info('Cannot find \'net\' on the PATH. Unable to configure DNS on host. Try manual configuration.')
+        return false
+      end
+
+      if `net start`.match(/Wired AutoConfig/m).nil?
+        info('Starting \'Wired AutoConfig\' service')
+        if self.class.admin_mode?
+          `net start dot3svc`
+        else
+          require 'win32ole'
```
budhrg: You should put it top as I found same step in the bottom of the this file.

lib/landrush/win_network_config.rb
```diff
@@ -0,0 +1,159 @@
...
+  class WinNetworkConfig
+    # Windows registry path under which network interface configuration is stored
+    INTERFACES = 'SYSTEM\\CurrentControlSet\\Services\\Tcpip\\Parameters\\Interfaces'.freeze
+
+    def initialize(env={})
+      @env = env
+    end
+
+    # Checks that all required tools are on the PATH and that the Wired AutoConfig service is started
+    def check_prerequisites
```
budhrg: `check_prerequisites` doesn't look like only checking it is starting some service also.

probably `check_and_start_prerequisites`?

lib/landrush/action/setup.rb
```diff
@@ -15,24 +15,35 @@ def call(env)
...
+        setup_host_resolver(env)
+      end
+
       def record_dependent_vm
         DependentVMs.add(machine_hostname)
       end
 
-      def setup_host_resolver
-        ResolverConfig.new(env).ensure_config_exists!
+      def setup_host_resolver(env)
```
budhrg: No need for linux?

budhrg:
@hferentschik Will check functionally tomorrow.

Gemfile.lock
hferentschik: slightly different gems between OS types. Makes no sense anymore to put it under version control :-(

lib/landrush/win_network_config.rb
hferentschik: Not what I have been talking about recently, but yes it is a "trick". It's a way to give an additional entry point into the code for when you need to run with admin privileges and/or try to spawn a process.

lib/landrush/win_network_config.rb
hferentschik: but here is is required. Trying to require win32ole on Linux will fail and throw LoadErrors. In this case we need to defer the loading until we know that we are on Windows. In the case we had the other day, it was a require for a plain gem afair.

lib/landrush/win_network_config.rb
hferentschik: Ahh, nice. I thought there was something ;-) Definitely going to change it. Did not like the fact anyways that I needed to make this a class method in the windows provisioning

lib/landrush/win_network_config.rb
hferentschik: you mean the require? If so, no. It needs to be dynamically loaded.

lib/landrush/win_network_config.rb
hferentschik: I'll call it _ensure_prerequisites_.

lib/landrush/action/setup.rb
hferentschik: There is need, but no code for it yet. Linux will continue to require manual configuration. Linux is also harder in a sense that we most likely have to differentiate between host types :-( OS X and Windows are easier in this regards. But adding support for Linux is definitely on the roadmap. Want to help!? ;-)

hferentschik:
@budhrg thanks for the code review. I'll apply the suggestions. I also find another issue during my tests on Windows 7 yesterday. Trying to resolve this this morning.

> Will check functionally tomorrow.

You might want to hold of until my next update.

hferentschik:
Closing this for now. Creating a new pull request.

Issue #234, opened 2016-08-05
sed command in RestartDnsmasq for redhat hosts does insert 127.0.0.1 before each nameserver entry


hferentschik:

hferentschik:
The problem is the used sed expression:

          SED_COMMAND = <<-EOF.gsub(/^ +/, '')
          sudo sed -i.orig '/nameserver/i\\
          nameserver 127.0.0.1 # Added by landrush, a vagrant plugin

          ' /etc/resolv.conf
          EOF

It does a global search and replace. The line should only be inserted once at the top. Something like this should work and be simpler:

     SED_COMMAND = "sudo sed -i.orig '1 i\\nameserver 127.0.0.1 # Added by landrush, a vagrant plugin' /etc/resolv.conf"

praveenkumar:
> Something like this should work and be simpler:

That expression gave below error

```
sed: -e expression #1, char 56: extra characters after command
```

What works is below expression. (Also tested.)

```
SED_COMMAND = <<-EOF.gsub(/^ +/, '') 
sudo sed -i.orig '1 i\
nameserver 127.0.0.1 # Added by landrush, a vagrant plugin
' /etc/resolv.conf
```

hferentschik:
@praveenkumar Can you try:

          SED_COMMAND = <<-EOF.gsub(/^ +/, '')
          sudo sed -i.orig '1 i\
          # Added by landrush, a vagrant plugin \
          nameserver 127.0.0.1 \
          ' /etc/resolv.conf
          EOF

praveenkumar:
> @praveenkumar Can you try:

This will give below result

```
# Added by landrush, a vagrant plugin           nameserver 127.0.0.1
```

hferentschik:
Needs to be \\

Can you try:

          SED_COMMAND = <<-EOF.gsub(/^ +/, '')
          sudo sed -i.orig '1 i\
          # Added by landrush, a vagrant plugin \\
          nameserver 127.0.0.1 \\
          ' /etc/resolv.conf
          EOF

hferentschik:
fixed by pull request #235

PR #235, opened 2016-08-05
Issue #234 Improving sed command used in RestartDnsmasq for redhat hosts

hferentschik:


lib/landrush/cap/host/redhat/restart_dnsmasq.rb
```diff
@@ -4,9 +4,9 @@ module Redhat
       module RestartDnsmasq
         class << self
           SED_COMMAND = <<-EOF.gsub(/^ +/, '')
-          sudo sed -i.orig '/nameserver/i\\
-          nameserver 127.0.0.1 # Added by landrush, a vagrant plugin
-
+          sudo sed -i.orig '1 i\
+          # Added by landrush, a vagrant plugin \\
+          nameserver 127.0.0.1 \\
```
praveenkumar: @hferentschik you don't need to put `\\` after `127.0.0.1` because it will add a blank line so 

```
sudo sed -i.orig '1 i\
# Added by landrush, a vagrant plugin \\
nameserver 127.0.0.1 
```
should work.

hferentschik: I kind of like the blank line, since it create a more visual separation as well :-)

Issue #241, opened 2016-08-06
Action::Teardown is not called for providers other than VirtualBox


hferentschik:

hferentschik:
fixed by pull request #242

PR #242, opened 2016-08-06
Issue #241 Appending teardown to generic hook to make sure it gets ca…

hferentschik:
…lled for all providers


LalatenduMohanty:
@hferentschik the commit message is very long. Can you please break into smaller statement.

PR #246, opened 2016-08-17
Read IP from guest even if there's just 1 private network

njam:
Fixes https://github.com/vagrant-landrush/landrush/issues/268

In case we want to use the public network's address.

For example my box has a private and a public address:
```
    config.vm.network :private_network, ip: '10.10.11.10'
    config.vm.network :public_network, :bridge => 'en0: Wi-Fi (AirPort)'
```

And I want to use the public one with:
```
config.landrush.host_interface = 'eth2'
```

For this we will need to use `read_host_visible_ip_address`.

@hferentschik wdyt?

### DIFF
--- a/lib/landrush/action/setup.rb
+++ b/lib/landrush/action/setup.rb
@@ -19,7 +19,11 @@ module Landrush
       end
 
       def host_ip_address
-        static_private_network_ip || machine.guest.capability(:read_host_visible_ip_address)
+        if private_network_ips.include? machine.config.landrush.host_ip_address
+          machine.config.landrush.host_ip_address
+        else
+          machine.guest.capability(:read_host_visible_ip_address)
+        end
       end
 
       private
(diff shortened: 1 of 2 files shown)


njam:
@hferentschik wdyt?

hferentschik:
So you want to skip the case in which the private network is automatically selected in case it is the sole private network, right? I guess this makes sense. I might have been a bit hasty when I added this "optimisation".

Do we have an actual issue for that and if not, do you mind creating one? At first glance this looks fine.

njam:
You mean to have a separate issue? Like this? https://github.com/vagrant-landrush/landrush/issues/268

Good to merge?

hferentschik:
> You mean to have a separate issue? Like this? #268 

Perfect :-)

hferentschik:
> Good to merge?

Merged, thanks :-)

PR #247, opened 2016-08-17
Change virtualbox DNS address to 10.0.2.3

njam:
> In that case the guest is assigned to the address 10.0.2.15, the gateway is set to 10.0.2.2 and the name server can be found at 10.0.2.3.
https://www.virtualbox.org/manual/ch09.html

Both work for me, but I think 10.0.2.3 is correct.

### DIFF
--- a/lib/landrush/action/redirect_dns.rb
+++ b/lib/landrush/action/redirect_dns.rb
@@ -24,7 +24,7 @@ module Landrush
       def _target_host
         case provider
         when :virtualbox then
-          '10.0.2.2'
+          '10.0.2.3'
         when :vmware_fusion, :libvirt then
           _gateway_for_ip(machine.guest.capability(:configured_dns_servers).first)
         when :parallels then


bexelbie:
@njam before I commit this, can you give me the backstory on what has led to this PR?  The fix seems benign, so I am thinking we should just commit it.

njam:
I just stumbled upon this, while working on something else.

I'm not sure why 10.0.2.2 works, but both addresses seem to work equally well:
```
# dig @10.0.2.2 -p 10053 www.cm.dev.cargomedia.ch
[…|
;; ANSWER SECTION:
www.cm.dev.cargomedia.ch. 0	IN	A	10.10.10.13

;; Query time: 7 msec
;; SERVER: 10.0.2.2#10053(10.0.2.2)
;; WHEN: Thu Aug 18 15:17:34 2016
;; MSG SIZE  rcvd: 58
```
```
# dig @10.0.2.3 -p 10053 www.cm.dev.cargomedia.ch
[…]
;; ANSWER SECTION:
www.cm.dev.cargomedia.ch. 0	IN	A	10.10.10.13

;; Query time: 7 msec
;; SERVER: 10.0.2.3#10053(10.0.2.3)
;; WHEN: Thu Aug 18 15:17:39 2016
;; MSG SIZE  rcvd: 58
```

bexelbie:
Cool.  I am planning to merge #249 on Monday/Tuesday and prep a release.  If this isn't already merged, I will merge it at that time too.

bexelbie:
Merged as 1df24371fc98f69f045d054a692be38d2330b32e

bexelbie:
Reverting as @hferentschik would like to discuss this further.

lib/landrush/action/redirect_dns.rb
```diff
@@ -24,7 +24,7 @@ def redirect_dns
       def _target_host
         case provider
         when :virtualbox then
-          '10.0.2.2'
```
hferentschik: Hmm, what do we really want here? The IP of the gateway to the host or the IP of the nameserver? Is it not the latter? And if so, is _10.0.2.2_ not the correct IP then?

njam: It's good that you ask, I'm not sure 100% about this change!

Docs:
> In NAT mode, the guest network interface is assigned to the IPv4 range 10.0.x.0/24 by default where x corresponds to the instance of the NAT interface +2. So x is 2 when there is only one NAT instance active. In that case the guest is assigned to the address 10.0.2.15, the gateway is set to 10.0.2.2 and the name server can be found at 10.0.2.3.
https://www.virtualbox.org/manual/ch09.html#changenat

So I *think* `10.0.2.3` is correct. But we can also leave `10.0.2.2` as nobody complained so far, and it worked just fine?

hferentschik: Looking at the other cases, it seems we are after the gateway here. The IP to reach the host. That's imo 10.0.2.2. I am tempted to leave it the way it is, since it works. I have not seen any actual errors around this. WDYT?

njam:
agree

> we can leave 10.0.2.2 as nobody complained so far, and it worked just fine

Issue #262, opened 2016-09-22
Add command to clear all entries at once


ezekg:
I've had multiple instances where a virtual machine fails somehow (be it testing provisioning scripts, random errors, etc.) causing Landrush to leave hanging entries. Then in order to work with another virtual machine using the same tld/ip, I have to manually remove all of Landrush's entries.

Currently, you're forced to run `vagrant landrush ls`, grab the top-most entry and then `vagrant landrush rm $entry` tens of times to clean it all out.

Could we add a `vagrant landrush clean` command or something similar?

I'll do the PR if you don't have time.

hferentschik:
Thanks, I like the idea. I have run into the "stale entries" issue myself several times. I think I'd prefer a flag for the `tm` command though. WDYT?

    landrush rm --all

That said, did you know you can at any stage just edit the configuration file which backs the host config. You find it in _~/.vagrant.d/data/landrush/hosts.json_. You can edit it with your favorite editor in order to clean it up or even delete it altogether between restarts. That's a bit easier than using `landrush rm` with a lot of entries.

> I'll do the PR if you don't have time.

If you have time, please send a pull request. Contributions are always welcome.

ezekg:
Great. I like `landrush rm --all` better as well. I didn't think to look for a data file anywhere, so that's good to know. I'll work on putting together a pull request soon.

hferentschik:
Fixed by pull request #287

ezekg:
Hey, thanks for wrapping this up! Sorry I wasn't able to get to it.

PR #266, opened 2016-09-23
Issue #262 Add --all flag to rm command

ezekg:
This will clear all Landrush host entries, resolving the issue of stale records building up over time as discussed in #262. There was already a `Store#clear!` method, so went ahead and utilized that.

## Usage
```bash
vagrant landrush rm --all
```

## Notes
- There didn't seem to be an existing implementation for command line flags, so if you'd like to do something different/more extensible, let me know.

### DIFF
--- a/lib/landrush/command.rb
+++ b/lib/landrush/command.rb
@@ -29,7 +29,11 @@ module Landrush
         Landrush::Store.hosts.set(host, ip)
       elsif command == 'del' || command == 'rm'
         key = ARGV[1]
-        Landrush::Store.hosts.delete(key)
+        if key == "--all"
+          Landrush::Store.hosts.clear!
+        else
+          Landrush::Store.hosts.delete(key)
+        end
       elsif command == 'help'
         @env.ui.info(help)
       else
@@ -63,7 +67,7 @@ module Landrush
         set { <host> <ip> | <alias> <host> }
           adds the given host-to-ip or alias-to-hostname mapping.
           existing host ip addresses will be overwritten
-        rm, del { <host> | <alias> }
+        rm, del { <host> | <alias> | --all }
           delete the given hostname or alias from the server
         help
           you're lookin at it!


ezekg:
Looks like the `Command` class may need a little bit of refactoring. 🚑

hferentschik:
Seems some Rubocop violations are the reason for the test failures. Any chance you can massage the code it a bit more? Otherwise I can see what I can do. Overall you are doing the right thing.

lib/landrush/command.rb
```diff
@@ -63,7 +67,7 @@ def help; <<-EOS.gsub(/^      /, '')
         set { <host> <ip> | <alias> <host> }
           adds the given host-to-ip or alias-to-hostname mapping.
           existing host ip addresses will be overwritten
-        rm, del { <host> | <alias> }
+        rm, del { <host> | <alias> | --all }
```
hferentschik: I think it should be more:

    rm, del [---all] { <host> | <alias> }  

WDYT?

ezekg:
Yeah, no problem. I'll see if I can refactor the `Command` class a bit when I have some free time.

lib/landrush/command.rb
ezekg: I was actually going to do that in my initial commit, but after thinking about it more, that would make `landrush rm --all <host>` valid per the `help` page, but `landrush rm <host> --all` would be invalid, which is kind of confusing.

Not to mention, the host being there is kind of irrelevant, since the `--all` flag is what we actually care about and will be checking for. What you suggested makes it look like a host/alias is required with an optional `--all` flag, which isn't the case.

What we're looking for is actually either or, not both. Right?

hferentschik: > What we're looking for is actually either or, not both. Right?

You are right. It makes sense as you describe it. I guess it looks a bit unusual.

hferentschik:
@ezekg I created pull request #287 which is based on your work, but addresses the rubocop issues and adding some tests. I hope this is ok with you.

PR #269, opened 2016-10-02
Update landrush ip dependency

Werelds:


hferentschik:
Fixes issue #271. Rebased and merged.

Issue #274, opened 2016-10-07
Host DNS configuration should consider NetworkManager on Fedora


jmontleon:
If you have dns=dnsmasq in the [main] section of /etc/NetworkManager/NetworkManager.conf landrush does not play well.

It should write the config to /etc/NetworkManager/dnsmasq.d rather than /etc/dnsmasq.d and restart (maybe reload?)  NetworkManager rather than dnsmasq as NetworkManager already manages a dnsmasq process on 127.0.0.1.

hferentschik:
@jmontleon, thanks for your feedback. Much appreciated. There were other reports around problems on Fedora, but yours has so far the most concrete solution. 

Mind you, the automatic host DNS configuration is really quite new (it only used to for for OS X out of the box) and given the wide variety of operation systems and how they are configured, it is hard to get this part right, especially if oneself is not using say Fedora as main development platform.

That said, you can always set _config.landrush.host_redirect_dns = false_ to false and prevent Landrush to do any sort of host configuration. You can then do the setup yourself. 

If your suggested solution works, maybe you would be interested in providing a pull request ;-)?

Last, but not least, what is the Fedora version you are running and do you know whether what you suggest will work for any version?

hferentschik:
Most likely a duplicate of #260

JPvRiel:
Hi, just to add a lesson learnt related to this:

- Even after doing a workaround with network manager, landrush causes dnsmasq to be installed if you forget `config.landrush.host_redirect_dns = false` in the Vagrantfile.
- Watch out for dnsmasq distro package update file default config - it can end up flooding syslog.

For context, a while ago I manually setup the dnsmasq settings landrush needs via NetworkManager's dnsmasq

```
$ sudo sh -c 'echo "server=/vagrant.test/127.0.0.1#10053" > /etc/NetworkManager/dnsmasq.d/vagrant-landrush'
$ sudo systemctl restart NetworkManager
$ grep '127.0.0.1#10053' /var/log/syslog | tail -n 1
Sep 27 00:05:47 <myhostname> dnsmasq[17285]: using nameserver 127.0.0.1#10053 for domain vagrant.test
```

But at some point I must have forgotten to put `config.landrush.host_redirect_dns = false` in one of my Vagrantfiles. The result:

- dnsmasq got installed
- `/etc/dnsmasq.d/vagrant-landrush-vagrant.test` got created

[...]

Issue #260, opened 2016-09-06
On Fedora register host DNS with NetworkManager


bexelbie:
Right now NetworkManager rewrites /etc/resolv.conf and overwrites our entry when there is a network change (for example from work to home networks).

hferentschik:
Any pointers on how to do this?

hferentschik:
Duplicate of issue #274

PR #275, opened 2016-10-08
Issue #274 Better support for dnsmasq run by NetworkManager

jmontleon:


lib/landrush/util/dnsmasq.rb
```diff
@@ -0,0 +1,10 @@
+module Landrush
+  module Util
+    class Dnsmasq
+      def self.nm_managed?(opts=nil)
+        nm_config = Pathname('/etc/NetworkManager/NetworkManager.conf')
+        File.exists?(nm_config) && File.readlines(nm_config).grep(/^dns=dnsmasq$/).any?
```
jmontleon: I could probably use the inifile rubygem to do a little better job parsing the file to ensure the line is in the main section, but not sure how you feel about adding another dependency for the sake of improving a single config check.

hferentschik:
Fix for issue #274

hferentschik:
Thanks for the pull request. I'll take a look asap.

jwmatthews:
Hello,  is there anything we can do to help this get merged and available in a release?

hferentschik:
>  is there anything we can do to help this get merged and available in a release?

It will most likely be in the next release. Just doing some more testing.

lib/landrush/util/dnsmasq.rb
```diff
@@ -0,0 +1,10 @@
+module Landrush
+  module Util
+    class Dnsmasq
+      def self.nm_managed?
+        nm_config = Pathname('/etc/NetworkManager/NetworkManager.conf')
+        File.exist?(nm_config) && File.readlines(nm_config).grep(/^dns=dnsmasq$/).any?
```
praveenkumar: I am using fedora 24 and when I grep dns=dnsmasq it doesn't have that string by default. I do have dnsmasq service up and running.
 
```
# cat /etc/NetworkManager/NetworkManager.conf 
# Configuration file for NetworkManager.
#
# See "man 5 NetworkManager.conf" for details.
#
# The directory /usr/lib/NetworkManager/conf.d/ can contain additional configuration
# snippets installed by packages. These files are read before NetworkManager.conf
# and have thus lowest priority.
# The directory /etc/NetworkManager/conf.d/ can contain additional configuration
# snippets. Those snippets override the settings from this main file.
#
# The files within one conf.d/ directory are read in asciibetical order.
#
# If /etc/NetworkManager/conf.d/ contains a file with the same name as
# /usr/lib/NetworkManager/conf.d/, the latter file is shadowed and thus ignored.
# Hence, to disable loading a file from /usr/lib/NetworkManager/conf.d/ you can
# put an empty file with the same name.
#
# If two files define the same key, the one that is read afterwards will overwrite
# the previous one.

[main]
#plugins=ifcfg-rh,ibft


[logging]
# When debugging NetworkManager, enabling debug logging is of great help.
#
# Logfiles contain no passwords and little sensitive information. But please
# check before posting the file online. You can also personally hand over the
# logfile to a NM developer to treat it confidential. Meet us on #nm on freenode.
# Please post full logfiles except minimal modifications of private data.
#
# You can also change the log-level at runtime via
#   $ nmcli general logging level TRACE domains ALL
# However, usually it's cleaner to enable debug logging
# in the configuration and restart NetworkManager so that
# debug logging is enabled from the start.
#
# You will find the logfiles in syslog, for example via
#   $ journalctl -u NetworkManager
#
#level=TRACE
#domains=ALL

# cat /etc/NetworkManager/NetworkManager.conf  | grep dnsmasq

$ systemctl status dnsmasq.service 
● dnsmasq.service - DNS caching server.
   Loaded: loaded (/usr/lib/systemd/system/dnsmasq.service; disabled; vendor preset: disabled)
   Active: active (running) since Thu 2016-11-03 16:25:32 IST; 21min ago
 Main PID: 8099 (dnsmasq)
    Tasks: 1 (limit: 512)
   CGroup: /system.slice/dnsmasq.service
           └─8099 /usr/sbin/dnsmasq -k
```

hferentschik: @jmontleon regarding this check for 'dns=dnsmasq' - is dnsmasq something you enabled manually (maybe for other purposes) prior to using Landrush or is this enabled by default on your Fedora box? Which version of Fedora are you using btw.

And if dnsmasq is not enabled for default, should we be enabling via NetworkManager by effectively adding _dns=dnsmasq_? See also issue #260?

hferentschik:
@jmontleon Did you get a chance to take another look at this?

lib/landrush/util/dnsmasq.rb
jmontleon: Yes. I believe it has to be enabled manually. It doesn't trigger the dnsmasq service to run as far as I know either; it runs under the NetworkManager service process, not to be confused with the dnsmasq processes that may be run by libvirt in a similar manner.

hferentschik:
Merged.

PR #294, opened 2016-12-04
Issue #297 Add Arch Linux host support

### LINKED ISSUE #297: Support Arch Linux as host
Add "caps" to install dnsmasq with *pacman*.

njam:
Closes #297

### DIFF
--- a/lib/landrush/cap/host/arch/dnsmasq_installed.rb
+++ b/lib/landrush/cap/host/arch/dnsmasq_installed.rb
@@ -1,0 +1,11 @@
+module Landrush
+  module Cap
+    module Arch
+      module DnsmasqInstalled
+        def self.dnsmasq_installed(_env, *_args)
+          system('pacman -Q dnsmasq > /dev/null 2>&1')
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


hferentschik:
Nice, do we have an issue for this as well? I think there is an issue with the Travis CI config as well. I have an idea what it is. Looking into it.

njam:
The problem is that we rely on vagrant 1.8.6 which requires bundler 1.12.5, but travis now has 1.13.6.
I think the best resolution is to [upgrade to vagrant 1.9.0](https://github.com/vagrant-landrush/landrush/pull/296) which removes the bundler requirement completely.

hferentschik:
> The problem is that we rely on vagrant 1.8.6 which requires bundler 1.12.5, but travis now has 1.13.6.

Right, I fixed that. 

> I think the best resolution is to upgrade to vagrant 1.9.0 which removes the bundler requirement completely.

Sure. Let's do that after the release. Current trunk contains several changes which are useful and it also takes users up to Vagrant 1.8.6. Then as you say, for Vagrant 1.9 we need a new plan. TBH, I have not yet looked into the changes Vagrant 1.9 brings. On my todo list. I guess I will start with your pull requests ;-)

njam:
Agree!

The build succeeds now.
I think this should be ready.

hferentschik:
Merged, thanks (squashed the commits)

Issue #295, opened 2016-12-04
Support Vagrant 1.9.0


njam:
- There's [no more bundler](https://github.com/mitchellh/vagrant/commit/1fb4553d14b235957d89d29f26f1c0decc971831) dependency! :dancer: 
- Gems are now installed in `~/.vagrant.d/gems/2.2.5/gems` instead of `~/.vagrant.d/gems/gems`. This breaks `start_server.rb`.

Also some problems reported here: https://github.com/vagrant-landrush/landrush/issues/292

hferentschik:
Thanks for looking into this. It was on my todo list, after seeing some reports with problems using 1.9.0. 

I think, however, I will still release a v1.2.0 with the current head (which will support Vagrant up to 1.9.0) and then we can address this issue.

khaije1:
A work-around (intended to follow version changes, so might not break right away) :)
```bash
ln -f -s $(find ~/.vagrant.d/gems/* -type d -maxdepth 0 2>/dev/null | sort -r | head -1) ~/.vagrant/gems/gems
```

joelio:
This workaround sorts it for me on OSX  (extended from @khaije1 note)

`ln -f -s $(find ~/.vagrant.d/gems/* -type d -maxdepth 0 2>/dev/null | sort -r | head -1) ~/.vagrant.d/gems/gems && vagrant plugin expunge --reinstall`

tristanbes:
any update on this ? we're using 1.9.1 but with the patch provided by @joelio , we're having `ERR_NAME_NOT_RESOLVED ` problems.

tristanbes:
Here's a working fix on OSX : 

```
cd ~/.vagrant.d/gems
ln -s 2.2.5/gems
```
do this BEFORE install landrush. Or uninstall the plugin, run the patch, and install again.

thanks to @nervo for sharing.

thor:
I'm no big fan of bumping things, so believe me when I say that I feel guilty for poking around here. :) With that said, I'm an avid landrush user, though I'm no good of a Ruby developer, so I'm merely interested in querying so that I can look into alternative solutions for myself.

Is it a matter of landrush currently being a deprioritized project, @hferentschik? It seems there are some forks with the vagrant-1.9 patches applied, as well as other minor things. #292 and #296 put a stick or so in the wheels. I wonder how come the issue is labeled *os/windows*, and not Linux and Mac as well, or just OS independent? :)

[...]

PR #296, opened 2016-12-04
Issue #295 Vagrant 1.9 compatibility

njam:
Addressing #295

Instead of running the DNS server with `ruby start_server.rb 10053 --gem_path=~/.vagrant.d/gems/gems`  we could start it as a plugin command:
```
vagrant landrush run-server 10053
```

This way we can avoid passing the gems directory along, because vagrant sets all of that up for us.

wdyt?


hferentschik:
> This way we can avoid passing the gems directory along, because vagrant sets all of that up for us.

Interesting idea. I need to look a bit closer at this. There seems to be some issues with the tests on Windows though.

njam:
It seems like `vagrant` is not on the PATH (is it the right term?) on this [Windows build](https://ci.appveyor.com/project/hferentschik/landrush-3agrx/build/24). Do you know what is the best course of action?

hferentschik:
> Do you know what is the best course of action?

Not out of the top of my head. If it is really just a PATH issue, we can just make sure vagrant is added to the PATH. One would expect though that it is. I guess I would like to know why it is not on the PATH first. Give me a few days to look into this. I am a bit swamped with other things right now, but hopefully I get a bit of time over the next days to look into this.

thor:
Anyone mind if I poke this issue to see what its status is? :)

robertoschwald:
Any news on this? Is this repo maintained anymore?

ezekg:
I'm embedding Landrush/Vagrant into a desktop app (essentially a GUI for Vagrant) and ended up having to include this script which runs after updating Vagrant to latest:

```bash
#!/bin/bash

# Fix compatibility issues with Vagrant >= 1.9.2 and Landrush 1.2.0
if [[ -L ~/.vagrant.d/gems/gems && -d ~/.vagrant.d/gems/gems ]]; then
  echo "already symlinked"
else
  echo "creating symlink"
  mkdir -p ~/.vagrant.d/gems/2.2.5/gems
  mkdir -p ~/.vagrant.d/gems/gems
  mv -f ~/.vagrant.d/gems/gems/* ~/.vagrant.d/gems/2.2.5/gems/ 2>&1 > /dev/null
  rm -r ~/.vagrant.d/gems/gems
  ln -sf ~/.vagrant.d/gems/2.2.5/gems ~/.vagrant.d/gems/gems 2>&1 > /dev/null
  echo "symlink created"
fi

# Install required plugins (just to be safe)
vagrant plugin install landrush --plugin-version 1.2.0
# … other plugins

# Attempt to repair existing plugins
vagrant plugin expunge --reinstall --force
```

The app also includes an embedded Ruby (2.2.5) so you may need to change that if you're running a different version of Ruby locally.

hferentschik:
I am going to close this pull request. I am not generally opposed to this, but as part of issue #324 Landrush already upgraded to Vagrant 2.1.2. AFAICT everything works fine. 

I can see that the proposed change makes things a bit easier since we let Vagrant deal with setting up the various paths. At the same time, it relies on vagrant being on the PATH. In most cases that might be the case, but not always (see the AppVeyor).

Regarding the problem which occurred when the gem path changed, the server start script worked as expected. However, due to the change one either needed to delete _~/.vagrant.d_ or do a reinstall of the plugin.

Issue #298, opened 2016-12-06
Travis CI used wrong Bundler version


hferentschik:
Bundler could not find compatible versions for gem "bundler":
    
      In Gemfile:
    
        vagrant was resolved to 1.8.6, which depends on
    
          bundler (= 1.12.5)
    
      Current Bundler version:
    
        bundler (1.13.6)
    
    This Gemfile requires a different version of Bundler.

PR #299, opened 2016-12-06
Issue #298 Updating Travis config to ensure that the right version of…

hferentschik:
… Bundler is installed. Also making sure that Rake init created build/vagrant.d directory


hferentschik:
Fix for issue #298

Issue #310, opened 2017-03-17
wired_autoconfig_service_running does not recognize running service on non-english windows platform


Valgard:
in my german windows 10 landrush does not recognize the running 'Wired AutoConfig' service:

[landrush] Using eth1 (x.x.x.x)
[landrush] starting 'Wired AutoConfig' service
Systemfehler 5 aufgetreten.

Zugriff verweigert

[landrush] Unable to start 'Wired AutoConfig' service. Unable to configure DNS on host. Try manual configuration.

a1ee9b:
I ran into the same problem. Try running Powershell or Cygwin as administrator.

tomcbe:
I ran into the same problem. I created a pull request to address this issue.

hferentschik:
Merged via pull request #326

PR #317, opened 2017-12-06
Fix method wired_autoconfig_service_running for windows

tomcbe:
Use locale independent variant to check if service dot3svc is running: sc query dot3svc
This fixes issue #310


hferentschik:
Superseded by pull request #326 which amend the approach.

PR #322, opened 2018-07-25
Change from OUTPUT to PREROUTING

ianmiell:
I believe PREROUTING was the correct iptables stage to use, and a kernel change causes problems on eg 18.04 ubuntu.

Tested on older kernels, and is back-compatible.

### DIFF
--- a/lib/landrush/cap/guest/linux/redirect_dns.rb
+++ b/lib/landrush/cap/guest/linux/redirect_dns.rb
@@ -15,7 +15,7 @@ module Landrush
         end
 
         def self._redirect_dns_rule(protocol, original_server, target_server, target_port)
-          "OUTPUT -t nat -p #{protocol} -d #{original_server} --dport 53 -j DNAT --to-destination #{target_server}:#{target_port}"
+          "PREROUTING -t nat -p #{protocol} -d #{original_server} --dport 53 -j DNAT --to-destination #{target_server}:#{target_port}"
         end
       end
     end


hferentschik:
@ianmiell sorry for the long turn-around. I was wondering how you tested the change. Changing from OUTPUT to PREROUTING does indeed prevent the occurrence of an error, but does this work. When I test locally with an Ubuntu 18.04 switching to PREROUTING, DNS does not work as expected anymore. For example, I am adding the following to my _Vagrantfile_:

    config.landrush.host 'static1.example.com', '1.2.3.4'
    config.landrush.host 'static2.example.com', '2.3.4.5'

Then once the machine is up, I ssh into it and _dig static1.example.com_. Does that work for you?

hferentschik:
So here is what I tried. Using the following Vagrantfile:

```Ruby
Vagrant.configure(2) do |config|
  config.vm.box = 'ubuntu/xenial64'
  #config.vm.box = 'ubuntu/bionic64'

  config.vm.hostname = 'foo.vagrant.test'

  config.landrush.host 'static1.example.com', '1.2.3.4'
  config.landrush.host 'static2.example.com', '2.3.4.5'
end
```

    $ vagrant up
    $ vagrant ssh -- dig static1.example.com

The latter should return the IP 1.2.3.4. It works with xenial64. It fails with bionic64 with the error:

    ../../../../lib/isc/unix/socket.c:2135: internal_send: 127.0.0.53#53: Invalid argument

Using the proposed changes to the iptables rules the dig command does not work at all anymore. For all I can the iptables does not get triggered in this case. One can, for example, look at the package count per rule using something like `sudo iptables -L OUTPUT -t nat  -v`.

IMO the iptables rules are sounds as they are. It looks more like an Ubuntu specific problem related to its current way to handle DNS.

Issue #321, opened 2018-07-19
iptables rule - move from OUTPUT => PREROUTING


ianmiell:
I had an issue with iptables on linux, moving from ubuntu 16.04 to 18.04.

It seems that there was a kernel change between the two versions (17.xx) which affects the OUTPUT chain, which now fails to forward properly. I can't find proof of this.

From experimenting it seems that changing OUTPUT to PREROUTING works on both 16.04 and 18.04, so is a change as simple as that likely to be accepted?

See also here, where I originally raised this with the vagrant image maintainer before looking back at landrush:

https://github.com/chef/bento/issues/1086

hferentschik:
Just so that we are on the same line, what is the error you are seeing? I get a:

      ../../../../lib/isc/unix/socket.c:2135: internal_send: 127.0.0.53#53: Invalid argument

I am not so sure whether changing OUTPUT to PREROUTING is going to work. My understanding is that locally generated packages flow through OUTPUT and not through PREROUTING. See also https://www.karlrupp.net/en/computer/nat_tutorial:

> For locally generated packets there is a small difference: Instead of passing through the PREROUTING chain it passes the OUTPUT chain and then moves on to the POSTROUTING chain.

I could be wrong, but I think with the suggested change the redirection of the DNS traffic just does not occur and hence the error does not get triggered. 

I am wondering whether the issue is related to Ubuntu's switch to _systemd-resolved_? However, I even get this error when I stop _systemd-resolved_. Not quite sure what's going on tbh.

ianmiell:
If memory serves, I tested by using a script that creates a VM with landrush. When running 18.04 networking doesn't work if landrush is used. When I changed the rule to PREROUTING all works fine. I also tried this in 16.xx and it continued to work.

I'll try and get a more detailed repro later.

ianmiell:
Yes, you're right. I can't figure it out, as I can telnet direct to the landrush port from the VM, and the iptables rule looks correct, and I can dig via that IP/port:

```

root@foo:~# iptables -L -t nat
Chain PREROUTING (policy ACCEPT)
target     prot opt source               destination

Chain INPUT (policy ACCEPT)
target     prot opt source               destination

Chain OUTPUT (policy ACCEPT)
target     prot opt source               destination
DNAT       tcp  --  anywhere             127.0.0.53           tcp dpt:domain to:10.0.2.2:10053
[...]

PR #336, opened 2018-11-28
Issue #321 Adding enable_route_localnet guest capability

hferentschik:
- Aligning some info messages format

Fix for issue #321

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -81,26 +81,6 @@ $ git merge fix-123
 $ git push origin master
 -------------------------------------------------------------------------
 
-== Releasing
-
-* Make sure the HEAD of development builds and passes all tests:
-  `bundle exec rake clean rubocop test features install`.
-* Review https://github.com/vagrant-landrush/landrush/milestones[milestone] for the current release
-** Make sure all issues are resolved
-** Move unresolved issues into an upcoming release
-** Close milestone
-* Update the link:CHANGELOG.md[CHANGELOG]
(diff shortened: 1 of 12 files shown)


hferentschik:
@ianmiell, what do you think about this? I had a crack at this issue. I don't think one really has to use the capabilities mechanism here, but I wanted to stick to the theme. WDYT?

ianmiell:
I'm afraid I ran out of time, and now work somewhere where I've not got
permission to contribute (and have less spare time). If memory serves I
thought caps might be a bit overkilll.

hferentschik:
> I'm afraid I ran out of time, and now work somewhere where I've not got
> permission to contribute (and have less spare time).

no worries

> If memory serves I thought caps might be a bit overkilll

Sure, it is probably not required. Definitely works without as well. I am a bit undecided on this one. The main reason to go with this approach was to stick with the existing structure of the code. As it stands now, the existing linux capabilities could be combined. The advantage would probably be a more compact and less indirect code. I don't think that using capabilities is otherwise a performance issues or similar. The advantage of the capabilities is that in the case one needs at some stage to differentiate  for any of these actions between different Linux distributions, one can easily do so. 

Given that the current pull request works, I am wondering whether it makes sense to merge it. Then in a follow up issue one could decide whether one would like to combine the existing Linux capabilities.

tristanbes:
Hello,

Since this patch is required to be able to use landrush with 18.04, is that possible you tag a version please ? @hferentschik

PR #345, opened 2019-04-17
Fixes #344 - uses registry instead of netsh

### LINKED ISSUE #344: Unable to determine network interface on non-English Windows
When using a Windows with a language different than English, `netsh` output will not be in English.

`lib/landrush/cap/host/windows/configure_visibility_on_host.rb` relies on `netsh` output being in English for it to correctly set the DNS server.

davividal:
When using non-English Windows, netsh output will not be in English.
Using Windows registry is locale independent.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -10,12 +10,6 @@
 Metrics/AbcSize:
   Max: 47
 
-# Offense count: 13
-# Configuration parameters: CountComments, ExcludedMethods.
-# ExcludedMethods: refine
-Metrics/BlockLength:
-  Max: 89
-
 # Offense count: 3
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
--- a/lib/landrush/cap/host/windows/configure_visibility_on_host.rb
+++ b/lib/landrush/cap/host/windows/configure_visibility_on_host.rb
@@ -1,5 +1,6 @@
 # This file needs to be able to execute out of the Vagrant context. Do not require any non core or non relative files
 require 'ipaddr'
+require 'English'
 require_relative '../../../util/retry'
 
 module Landrush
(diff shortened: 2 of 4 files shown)


davividal:
This PR aims to fix #344 .

lib/landrush/cap/host/windows/configure_visibility_on_host.rb
```diff
@@ -1,5 +1,6 @@
 # This file needs to be able to execute out of the Vagrant context. Do not require any non core or non relative files
 require 'ipaddr'
+require 'English'
```
hferentschik: Why is that needed?

davividal: Because of this:

    lib/landrush/cap/host/windows/configure_visibility_on_host.rb:29:13: C: Style/SpecialGlobalVars: Prefer $CHILD_STATUS from the stdlib 'English' module (don't forget to require it) over $?.
            $?.exitstatus.zero?
            ^^

hferentschik: ahh, got you.

.rubocop_todo.yml
```diff
@@ -15,6 +15,12 @@ Metrics/AbcSize:
 # ExcludedMethods: refine
 Metrics/BlockLength:
   Max: 89
+  Exclude:
```
hferentschik: So instead of keep adding to the rubocop_todo config, why not try to settle on better defaults. Something like this in _rubocop.yml_ would work for me:

```yaml
Metrics/ModuleLength:
  Max: 200

Metrics/BlockLength:
  Max: 50
  ExcludedMethods: ['describe', 'context']
```

I find the default values to low and I don't think it is worth trying to refactor just to meet these defaults. A max module length of 200 and a max block length of 50 seems reasonable to me. 
Also imo it makes sense to exclude describe and context which would mainly match the minitests. 

If we go down this route, we can remove BlockLength and ModuleLength from the rubocop_todo. WDYT?

lib/landrush/cap/host/windows/configure_visibility_on_host.rb
hferentschik: ok

lib/landrush/cap/host/windows/configure_visibility_on_host.rb
```diff
@@ -19,46 +20,82 @@ def configure_visibility_on_host(env, ip, tlds)
...
           end
 
-          private
+          # Fetches IP/mask info from registry from a given registry entry
+          # This is a separate method to make testing easier
+          def interface_address(interface)
+            dhcp_enabled = interface.read('EnableDHCP')[1]
+
+            if dhcp_enabled
+              if_ip = interface.read('DhcpIPAddress')[1]
```
hferentschik: Interesting. Did you by any chance test this against a DHCP enabled Windows box as well as one with a static IP?

lib/landrush/cap/host/windows/configure_visibility_on_host.rb
```diff
@@ -19,46 +20,82 @@ def configure_visibility_on_host(env, ip, tlds)
...
+              if_mask = interface.read('DhcpSubnetMask')[1]
+            else
+              if_ip = interface.read('IPAddress')[1]
+              if_mask = interface.read('SubnetMask')[1]
+            end
+
+            if if_ip.is_a? Array
```
hferentschik: What are the elements of the array? Do you know in which cases this occurs?

lib/landrush/cap/host/windows/configure_visibility_on_host.rb
```diff
@@ -19,46 +20,82 @@ def configure_visibility_on_host(env, ip, tlds)
...
+              if_ip = if_ip[0]
```
hferentschik: So the first one is always the right one? Do you have some example interface config?

test/landrush/cap/host/windows/configure_visibility_on_host_test.rb
```diff
@@ -107,31 +65,119 @@ module Windows
           end
         end
 
-        describe '#get_network_name' do
-          it 'returns network name for matching IP' do
-            ConfigureVisibilityOnHost.expects(:`).with('netsh interface ip show config').returns(NETSH_EXAMPLE)
-            expect(ConfigureVisibilityOnHost.get_network_name('192.168.1.193')).must_equal('Ethernet')
+        describe '#get_network_guid' do
```
hferentschik: 👍 nice tests

hferentschik:
Good stuff :-)

lib/landrush/cap/host/windows/configure_visibility_on_host.rb
davividal: Yes, I did. :) That's, actually, how I came across this. :)

lib/landrush/cap/host/windows/configure_visibility_on_host.rb
davividal: From what I saw, this only occurs when you have DHCP enabled.
The elements are... a single IP.
I didn't find why Windows stores static IPs as a single-value and DHCP as multi-value in registry.

lib/landrush/cap/host/windows/configure_visibility_on_host.rb
davividal: Yes, here you have: https://gist.github.com/davividal/0518b75f9fb4fca86f29c6f33373f938

.rubocop_todo.yml
davividal: I think that lowering BlockLength (from 89 to 50) *and* excluding describe and context are a nicer solution.
For the ModuleLength I think that excluding tests makes more sense than increasing the limit.

davividal:
Yes, I tested this on one Windows box.

lib/landrush/cap/host/windows/configure_visibility_on_host.rb
hferentschik: Cool

lib/landrush/cap/host/windows/configure_visibility_on_host.rb
hferentschik: Hmm, a bit random, but fair enough I guess

.rubocop_todo.yml
```diff
@@ -14,7 +14,12 @@ Metrics/AbcSize:
 # Configuration parameters: CountComments, ExcludedMethods.
 # ExcludedMethods: refine
 Metrics/BlockLength:
-  Max: 89
+  Max: 50
```
hferentschik: Should we not move this into `.rubocop.yml`? The idea is to eventually get rid of `.rubocop_todo.yml` , either by refactoring code to not require any non-default configuration or by setting some reasonable project specific configuration in  `.rubocop.yml`.

.rubocop_todo.yml
```diff
@@ -14,7 +14,12 @@ Metrics/AbcSize:
...
+  ExcludedMethods: ['describe', 'context']
+
+Metrics/ModuleLength:
+  Exclude:
+    - !ruby/regexp /^test\//
```
hferentschik: Sweet. I did not know that one can do that.

lib/landrush/cap/host/windows/configure_visibility_on_host.rb
davividal: After reading around, it looks like this is like this for a reason: if the interface is not using DHCP, you can have multiple IPs on the same interface.
This makes sense to me. But I don't have a use case for this, so I am not able to properly test. I'm pretty sure that this (multiple IPs on the same NIC) will break things, but I think that this is such an edge case that I shouldn't care about it *right now*. :)

.rubocop_todo.yml
davividal: Moved.

davividal:
Just squashed the commits.

Issue #346, opened 2019-05-04
Upgrade Vagrant development version


hferentschik:
- Upgrade Vagrant dev version
- Disable failing Cucumber feature (see #342)
- Update maintainers list

hferentschik:
- Upgrade Vagrant dev version
- Disable failing Cucumber feature (see #342)

hferentschik:
Merged via pull request #347

Issue #342, opened 2019-02-20
landrush host visibility of vm  is not working on ubuntu 18.04


kundeng:
I installed it and I can get vms to see each other, and as well as the internet(e.g. www.google.com) but 

I can't seem to get my host to ping or resolve the vm's names.  

I'm new to this,  can someone explain what is wrong and how to fix it?  

I checked the following on the host:

dig -p 10053  orc-app1.test @localhost
and it works,  proving that the landrush dns at 10053 is working fine.  

I think my host is ubuntu 18.04 and it has that systemd-resolved stub listener running on port 53, so dnsmasq is not running at port 53.   Is this a problem?

coelhoricardo:
landrush is based on dnsmasq. but since 18.04, ubuntu replaced dnsmasq for systemd-resolved.

ericenns:
I got this working on Ubuntu 18.04 by changing my `/etc/systemd/resolved.conf` to have the following content:
```
#  This file is part of systemd.                                                                                                                                                                                                                                                                                              
#                                                                                                                                                                                                                                                                                                                             
#  systemd is free software; you can redistribute it and/or modify it                                                                                                                                                                                                                                                         
#  under the terms of the GNU Lesser General Public License as published by                                                                                                                                                                                                                                                   
#  the Free Software Foundation; either version 2.1 of the License, or                                                                                                                                                                                                                                                        
#  (at your option) any later version.                                                                                                                                                                                                                                                                                        
#                                                                                                                                                                                                                                                                                                                             
# Entries in this file show the compile time defaults.                                                                                                                                                                                                                                                                        
# You can change settings by editing this file.                                                                                                                                                                                                                                                                               
# Defaults can be restored by simply deleting this file.                                                                                                                                                                                                                                                                      
#                                                                                                                                                                                                                                                                                                                             
# See resolved.conf(5) for details                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                              
[Resolve]                                                                                                                                                                                                                                                                                                                     
DNS=127.0.0.1                                                                                                                                                                                                                                                                                                                 
#FallbackDNS=                                                                                                                                                                                                                                                                                                                 
#Domains=                                                                                                                                                                                                                                                                                                                     
#LLMNR=no                                                                                                                                                                                                                                                                                                                     
#MulticastDNS=no                                                                                                                                                                                                                                                                                                              
#DNSSEC=no                                                                                                                                                                                                                                                                                                                    
[...]

PR #347, opened 2019-05-04
Issue #346 Updating Vagrant dev version

hferentschik:
- disabling failing Cucumber test
- updating maintainers list

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -83,10 +83,6 @@ $ git push origin master
 
 == Maintainers
 
-* Brian Exelbierd (@bexelbie)
-* Eric Sorenson (@ahpook)
-* Florian Holzhauer (@fh)
+* Davi Koscianski Vidal (@davividal)
 * Hardy Ferentschik (@hferentschik)
-* Josef Strzibny (@strzibny)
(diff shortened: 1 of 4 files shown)


hferentschik:
@davividal wdyt?

davividal:
Looks nice for me.
I'm currently having issues with bento/debian boxes when running on Mac. I'm still trying to understand the issue so I can write a test for it and solve. So, I see no problems in disabling some failing tests since I'm bound to look into that in the (very) near future.

hferentschik:
> Looks nice for me.

Cool.

> I'm currently having issues with bento/debian boxes when running on Mac.

Anything which is reported? Which version of bento/debian? I am on macOS as well. 

> I'm still trying to understand the issue so I can write a test for it and solve. 

Sounds good. Regarding the disabled test. It used to work, but it seems something has changed. See also #342. 

The Cucumber tests are a great way to run through the most common use cases. Be aware, that if a Cucumber run fails you might have to manually check and kill a locally running Landrush process. I am looking into how to improve this. Probably by a _beforeEach_ hook. 

> So, I see no problems in disabling some failing tests since I'm bound to look into that in the (very) near future.

+1

PR #350, opened 2019-06-30
Update Usage.adoc

bish0polis:
It can be debated but I think we mean 'lose', here.


davividal:
Yes, you are correct. Thanks.

davividal (approved):

